PR #20, opened 2014-10-21
[FIXED JENKINS-20327][FIXED JENKINS-23221][FIXED JENKINS-24804][FIXED JE...

stephenc:
...NKINS-21064] java.lang.IllegalStateException: Form too large

The default should be `-1`, i.e. no limit, for maximum compatibility with the original winstone

### DIFF
--- a/src/java/winstone/Launcher.java
+++ b/src/java/winstone/Launcher.java
@@ -138,6 +138,8 @@ public class Launcher implements Runnable {
             if (maxParameterCount>0) {
                 server.setAttribute("org.eclipse.jetty.server.Request.maxFormKeys",maxParameterCount);
             }
+            server.setAttribute("org.eclipse.jetty.server.Request.maxFormContentSize",
+                    Option.REQUEST_FORM_CONTENT_SIZE.get(args));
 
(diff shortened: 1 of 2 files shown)


stephenc:
@reviewbybees (cleaning out my old PRs)

oleg-nenashev:
I'd rather put :bee: on the fix as is. 
Additional monitoring and warnings could be useful, but almost any Jenkins installation suffers from this issue at some point => admins setup their own random values.

daniel-beck:
I wonder whether defaulting to -1 is a sensible default on the level of Winstone-Jetty, or whether it shouldn't just be parameterized and then set to -1 in the config files bundled with the package installers.

WDYT?

jglick:
> whether it shouldn't just be parameterized

Is that not what `--requestFormContentSize` accomplishes?

Anyway :bee: from me. Would still be great to do automatic diagnosis of huge requests, but that is clearly more work. Maybe something for `support-core`?

stephenc:
@reviewbybees done

stephenc:
This PR has been sitting around for over a year... I'm merging

PR #22, opened 2016-02-19
[JENKINS-23378] Jetty9 upgrade

kohsuke:
As a part of upgrading Jenkins to servlet 3.1, we need Winstone to support servlet 3.1, which this PR does by pulling in Jetty 9.2, which is the latest version of Jetty that runs on Java7.

Notable incompatibilities include:
* Lack of AJP support ([link](https://github.com/eclipse/jetty.project/issues/134))
* requestBufferSize setting in Jetty8 appears to have no counterpart

This PR is to let others comment on the changes thus far, find those missing features from somewhere, and try it out on Jenkins. Once we decide to drop those functionalities, we need a follow-up change to clean up the help screen, really remove AJP related code, etc.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.jenkins-ci</groupId>
   <artifactId>winstone</artifactId>
-  <version>2.10-SNAPSHOT</version>
+  <version>3.0-SNAPSHOT</version>
   <packaging>jar</packaging>
 
   <name>Winstone</name>
--- a/pom.xml
+++ b/pom.xml
@@ -221,25 +221,27 @@
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-servlet</artifactId>
-      <version>8.1.13.v20130916</version>
+      <version>9.2.15.v20160210</version>
     </dependency>
 
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-webapp</artifactId>
-      <version>8.1.13.v20130916</version>
+      <version>9.2.15.v20160210</version>
     </dependency>
 
+    <!--  See https://github.com/eclipse/jetty.project/issues/134
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-ajp</artifactId>
-      <version>8.1.13.v20130916</version>
+      <version>9.2.15.v20160210</version>
     </dependency>
+    -->
 
     <dependency>
       <groupId>org.eclipse.jetty.spdy</groupId>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -221,25 +221,27 @@
...
+      <version>9.2.15.v20160210</version>
     </dependency>
+    -->
 
     <dependency>
       <groupId>org.eclipse.jetty.spdy</groupId>
-      <artifactId>spdy-jetty-http</artifactId>
-      <version>8.1.13.v20130916</version>
+      <artifactId>spdy-http-server</artifactId>
+      <version>9.2.15.v20160210</version>
```
jglick: Introduce a property for these so you need not repeat yourself.

src/java/winstone/Ajp13ConnectorFactory.java
```diff
@@ -30,6 +28,9 @@ public boolean start(Map args, Server server) throws IOException {
             return false;
         }
 
+        throw new UnsupportedOperationException();
```
jglick: Why not just delete the whole class?

src/java/winstone/cmdline/Option.java
```diff
@@ -111,6 +111,10 @@
     public static final OBoolean USAGE=bool("usage",false);
     public static final OInt SESSION_TIMEOUT=integer("sessionTimeout",-1);
     public static final OInt REQUEST_HEADER_SIZE=integer("requestHeaderSize",8192); // default for jetty 8
+    /**
+     * @deprecated
+     *      Doesn't appear to have the corresponding property in Jetty9
+     */
     public static final OInt REQUEST_BUFFER_SIZE=integer("requestBufferSize",16384); // default for jetty 8
```
jglick: So clearer to just delete it.

src/test/winstone/Ajp13ConnectorFactoryTest.java
```diff
@@ -27,7 +28,7 @@
  * @author Kohsuke Kawaguchi
  */
 public class Ajp13ConnectorFactoryTest extends AbstractWinstoneTest {
-    @Test
+    @Test @Ignore("Jetty9 doesn't support AJP yet")
```
jglick: Better just delete the file.

jglick:
:+1:

kohsuke:
I took a quick look at [jetty-ajp code](https://github.com/eclipse/jetty.project/tree/jetty-8/jetty-ajp). While the code is relatively small, the relevant abstractions have so drastically changed in Jetty 9 that the porting is not practical unless one knows the architecture of Jetty deeply.

So AJP support is not coming back.

src/java/winstone/Ajp13ConnectorFactory.java
kohsuke: Hopefully successive change made it clear why

PR #23, opened 2016-03-18
[JENKINS-25333] Self Signed certificates with JDK8

christ66:
@reviewbybees JDK 8 moved the `CertAndKeyGen` class to a new location.

This code change uses reflection to make make sure it is compatible with JDK 7 and JDK 8.

Unit tests pass in both JDK 8 and JDK 7 in my local environment:

screen shot 2016-03-17 at 10 30 30 pm
screen shot 2016-03-17 at 10 28 10 pm

### DIFF
--- a/src/java/winstone/HttpsConnectorFactory.java
+++ b/src/java/winstone/HttpsConnectorFactory.java
@@ -15,7 +15,7 @@ import org.eclipse.jetty.util.B64Code;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import sun.security.util.DerInputStream;
 import sun.security.util.DerValue;
-import sun.security.x509.CertAndKeyGen;
+//import sun.security.x509.CertAndKeyGen;
 import sun.security.x509.X500Name;
 import winstone.cmdline.Option;
 
@@ -27,6 +27,7 @@ import java.io.FileInputStream;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
+import java.lang.reflect.InvocationTargetException;
 import java.math.BigInteger;
 import java.security.GeneralSecurityException;
 import java.security.KeyFactory;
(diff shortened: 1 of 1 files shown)


aheritier:
:bee: but with JDK 9 out in few .... cycles we should perhaps start a more global cleanup to get ride of all usages of internal Java APIs like this. J9 won't allow us to access to them...

jglick:
I presume this is intending to address [JENKINS-25333](https://issues.jenkins-ci.org/browse/JENKINS-25333).

Better than nothing, I guess, but not really a fix. The right fix would be to use some supported API, not JRE internal classes which are subject to incompatible change (or outright deletion) without notice.

christ66:
@jglick Yes I forgot to post this fixes JENKINS-25333.

@aheritier @jglick Would a better solution be to use bouncycastle? I've been trying to investigate in the JDK if it's possible to generate a self-signed cert without requiring `sun.*` packages, but I'm not sure it's feasible.

aheritier:
In the past I didn't have a good experience with BouncyCastle and how they regularly broke the backward compatibility. ex:
https://jira.exoplatform.org/browse/CRASH-188
https://jira.exoplatform.org/browse/CRASH-228

But I agree that we don't have a lot of choices

smithj4:
This is now an issue for RedHat Enterprise 5, 6 & 7, since a few weeks ago, a java-1.7.0-openjdk security update was released for all RHEL versions that updated it to u111 and also moves the CertAndKeyGen class from /sun/security/x509/CertAndKeyGen.class to /sun/security/tools/keytool/CertAndKeyGen.class:

https://rhn.redhat.com/errata/RHSA-2016-1504.html

It looks like this comes directly from the upstream openjdk, from their Mercurial source code repo:

http://hg.openjdk.java.net/jdk7u/jdk7u/jdk/file/tip/src/share/classes/sun/security/tools/keytool/CertAndKeyGen.java

The necessity of getting this fixed soon is I hope a little more important now, since it also affects Java 7, not only 8.

jglick:
Can we just amend this to mark the option to generate a certificate as officially deprecated, both in the `--help` and as a runtime warning? And then here we are making a best effort to keep people running for a while until they fix their setup.

jtnord:
@jglick Well I gathered from the feedback that #26 is not currently merge-able.
Lets do this and we can trivially bump the JDK version after?

Currently winstone is not building for me and I would like to rule environmental issues vs its just not working on windows issues.

PR #24, opened 2016-04-28
[INFRA-588] Use new distributionManagement.url

jglick:
@reviewbybees & @batmat

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -172,7 +172,7 @@
   <distributionManagement>
     <repository>
       <id>maven.jenkins-ci.org</id>
-      <url>http://maven.jenkins-ci.org:8081/content/repositories/releases</url>
+      <url>https://repo.jenkins-ci.org/releases/</url>
     </repository>
   </distributionManagement>
 


pom.xml
```diff
@@ -172,7 +172,7 @@
   <distributionManagement>
     <repository>
       <id>maven.jenkins-ci.org</id>
-      <url>http://maven.jenkins-ci.org:8081/content/repositories/releases</url>
+      <url>https://repo.jenkins-ci.org/releases/</url>
```
recena: `jenkins-ci.org` or `jenkins.io`?

recena: It seems the valid url is `https://repo.jenkins.io/releases/` but I thought the idea was to move it to `repo.jenkins.io`.

Forget my comment.

recena:
🐝

pom.xml
jglick: I am just going by the recommended URL given by @batmat on the dev list.

batmat: I just had a chat with @rtyler about that to be sure. 
TL;DR. For technical reasons, using `.jenkins-ci.org` for now is fine. There's no short-term plan to switch over to `.jenkins.io` though @recena is right that that would be logical.

```
[22:31] <     rtyler> | we don't have a CNAME for repo.jenkins.io nor a certificate, given the shit-show of getting JFrog a certificate for repo.jenkins-ci.org, I'm don't think it would be very 
          easy to switch to .io
[22:31] <     rtyler> | batmat: I think that change is fine for now
```

batmat:
:+1:

PR #25, opened 2016-11-02
[FIXED JENKINS-39436] set/unset server version support

escoem:
please @reviewbybees cc. @aheritier @daniel-beck

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.ipr
 *.iws
 target
+.idea
+dependency-reduced-pom.xml
--- a/src/java/winstone/HttpConnectorFactory.java
+++ b/src/java/winstone/HttpConnectorFactory.java
@@ -42,6 +42,7 @@ public class HttpConnectorFactory implements ConnectorFactory {
             HttpConfiguration config = connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration();
             config.addCustomizer(new ForwardedRequestCustomizer());
             config.setRequestHeaderSize(Option.REQUEST_HEADER_SIZE.get(args));
+            config.setSendServerVersion(Option.SEND_SERVER_VERSION.get(args));
 
             server.addConnector(connector);
             return true;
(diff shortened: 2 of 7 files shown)


aheritier:
🐝

daniel-beck:
https://wiki.jenkins-ci.org/display/JENKINS/GitHub+commit+messages

daniel-beck:
The issue does not present a rationale for this change, making it impossible to judge the value of this change, so :-1: out of principle.

Depending on the reason for this change, maybe we don't need an option at all, and could just always set this flag? Again, impossible to evaluate because the ticket is useless.

escoem:
@daniel-beck I have updated the JIRA issue https://issues.jenkins-ci.org/browse/JENKINS-39436

This PR provides a configuration option for unset the HTTP response header.

Server: Jetty(9.2.z-SNAPSHOT)

amuniz:
@daniel-beck in general any security analysis will suggest to hide any information that could give opportunities to an attacker, and server name and version is a good clue for them.

Giving the option to hide the `Server` header is something a lot of people is doing by setting up another network element in front of Jenkins, but why should we require that? Specially if Jetty is already giving us the option to do it.

I'm :+1: on this change (and leave everything as it is now by default, so we avoid any possible backward issue).

daniel-beck:
@amuniz 

> hide any information that could give opportunities to an attacker, and server name and version is a good clue for them

In general, yes. But Jetty is, very publicly, the default container for Jenkins. And Jenkins is trivial to detect by its headers. How does this slow anyone down even a bit?

---

Does this at least remove the server signature from 404 error pages as wekk, or does it only take a single request to a non-existing resource to still see the following, despite this change?

> Powered by Jetty://

oleg-nenashev:
@amuniz @escoem this PR is waiting for your response

amuniz:
@daniel-beck @oleg-nenashev hiding the server is like a good security practice in general, I don't see a reason to not apply this (although I don't see it as something really important), so up to you.

daniel-beck:
@amuniz No response to the second half of my previous comment. If that's not also addressed, applying this would be a joke. Otherwise, we can (and could even default to false IMO, not opposed).

daniel-beck:
@escoem Merge conflict.

oleg-nenashev:
Added issue labels. @escoem @aheritier  any plans to work on it?

daniel-beck:
Nothing changed since my previous comments.

escoem:
@oleg-nenashev @aheritier @daniel-beck does fixing the merge conflict make sense?

oleg-nenashev:
I think this change is not going to solve any security concern in general. Security through obscurity, but OTOH I agree with @aheritier that it's a common practice. If the PR is finalized and confirmed to work properly, I am fine with this change.

PR #26, opened 2016-12-08
add a Jenkinsfile

jtnord:
@reviewbybees /cc @rtyler

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,54 @@
+#!/usr/bin/env groovy
+
+/* Only keep the 10 most recent builds. */
+properties([[$class: 'BuildDiscarderProperty',
+                strategy: [$class: 'LogRotator', numToKeepStr: '10']]])
+
+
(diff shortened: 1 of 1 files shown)


jtnord:
@rtyler you can not build this with JDK8 - the code used to build the self signed certificates was (re)moved with JDK8.   ([JENKINS-25333](https://issues.jenkins-ci.org/browse/JENKINS-25333))

rtyler:
That's fun. At some point in 2017, JDK7 support is going to likely disappear from Jenkins infrastructure entirely, as Jenkins itself moves away from it.

*shrugs*

jglick:
Then let us merge #23 first. I do not think we should be building anything on JDK 7 at this point.

jtnord:
@jglick Well I gathered from the feedback that #23 is not currently merge-able.
Lets do this and we can trivially bump the JDK version after?

Currently winstone is not building for me and I would like to rule environmental issues vs its just not working on windows issues.

PR #28, opened 2016-12-15
update ci build to use JDK8

jtnord:
Updates the Jenkinsfile to use JDK8 for the builds.

@reviewbybees   Blocked by #23

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ for (int i = 0; i < platforms.size(); ++i) {
 
                 stage('Build') {
                     withEnv([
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
I'd rather 🐛 it till Jenkins project drops Java 7. Just because it may capture some additional issue.

jglick:
@oleg-nenashev why? We run most component builds on JDK 8, and rely on `-target` and Animal Sniffer to detect issues.

oleg-nenashev:
@jglick But do we have Animal Sniffer in this component? From what I see "no"

jglick:
I will deal with that, just a minute…

PR #33, opened 2017-02-17
[FIXED JENKINS-42159] If startup fails, exit the process.

kohsuke:
As it stands right now, empty Jetty keeps running and makes it difficult
to detect catastrophic boot error.

The code change makes any exception thrown from ServletContextListener a
fatal problem.

https://issues.jenkins-ci.org/browse/JENKINS-42159

### DIFF
--- a/src/java/winstone/Launcher.java
+++ b/src/java/winstone/Launcher.java
@@ -352,6 +352,7 @@ public class Launcher implements Runnable {
             new Launcher(args);
         } catch (Throwable err) {
             Logger.log(Logger.ERROR, RESOURCES, "Launcher.ContainerStartupError", err);
+            System.exit(1);
         }
     }
     
--- a/src/java/winstone/HostConfiguration.java
+++ b/src/java/winstone/HostConfiguration.java
@@ -180,6 +180,7 @@ public class HostConfiguration {
             }
         };
         wac.getSecurityHandler().setLoginService(loginService);
+        wac.setThrowUnavailableOnStartupException(true);    // if boot fails, abort the process instead of letting empty Jetty run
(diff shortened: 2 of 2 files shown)


kohsuke:
@reviewbybees

src/java/winstone/Launcher.java
```diff
@@ -161,8 +161,10 @@ public Launcher(Map args) throws IOException {
 
             success = true;
         } finally {
-            if (!success)
+            if (!success) {
                 shutdown();
+                System.exit(1);
```
oleg-nenashev: Needs a message in the log

src/java/winstone/Launcher.java
```diff
@@ -352,6 +354,7 @@ public static void main(String argv[]) throws IOException {
             new Launcher(args);
         } catch (Throwable err) {
             Logger.log(Logger.ERROR, RESOURCES, "Launcher.ContainerStartupError", err);
```
oleg-nenashev: Is it FATAL/SEVERE now?

kohsuke: Note that this is Winstone's logger and this maps to jul's SEVERE level already

kohsuke:
Removed unnecessary `System.exit` in the original patch

src/java/winstone/Launcher.java
oleg-nenashev: right

oleg-nenashev:
@kohsuke The issue is raised as an "Improvement". Just in case, do you consider it as an LTS candidate?

kohsuke:
No, I don't think LTS backport is necessary.

PR #32, opened 2017-01-23
[JENKINS-40693, JENKINS-43713] upgrade to jetty 9.4.x

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>

see https://issues.jenkins-ci.org/browse/JENKINS-40693

PR #34, opened 2017-04-08
Bump java.level to 8 as jenkins core uses 8 now.

paladox:


oleg-nenashev:
Will integrate later once we get the last patch in #32 released and backported

olamy:
note it's already included in #32

oleg-nenashev:
Closing since #32 has been integrated

PR #38, opened 2017-06-08
README: Update documentation to reflect the Winstone 4 state

oleg-nenashev:
The previous documentation is just incorrect for the new release: we use Jetty 9.4 && do not support SPDY anymore.

@reviewbybees @olamy


oleg-nenashev:
@reviewbybees done
I will merge it since it seems to be a trivial follow-up to the changelog

PR #43, opened 2017-11-02
trivial README edit: delegated *to* Jetty

apetro:
Rather than

> the gut of Winstone has been removed and delegated Jetty

instead

> the gut of Winstone has been removed and delegated to Jetty


oleg-nenashev:
thanks!

PR #45, opened 2018-02-22
[JENKINS-49596] - Increase timeout to 30 min, Noting 4.1.1 and adding it to documentation

oleg-nenashev:
@olamy @dwnusbaum I have released 4.1.1 for testing on the Master's side. Actually it should have been a staged timestamped snapshot, but it was too late when I realised that. In the worst case we can mark release as burned one

For now I did some polishing: https://issues.jenkins-ci.org/browse/JENKINS-49596

@reviewbybees

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,17 @@ Changelog
 
 ##4.2
 
+Release date: Coming soon
+
 * Update Jetty from `9.4.5.v20170502` to `9.4.8.v20171121` [CHANGELOG](https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.8.v20171121)
-* User session memory leak. [JENKINS-49596](https://issues.jenkins-ci.org/browse/JENKINS-49596)
+
+## 4.1.1
+
+Release date: Feb 22, 2017
+
+* [JENKINS-49596](https://issues.jenkins-ci.org/browse/JENKINS-49596) -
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,8 +3,17 @@ Changelog
...
+
 * Update Jetty from `9.4.5.v20170502` to `9.4.8.v20171121` [CHANGELOG](https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.8.v20171121)
-* User session memory leak. [JENKINS-49596](https://issues.jenkins-ci.org/browse/JENKINS-49596)
+
+## 4.1.1
+
+Release date: Feb 22, 2017
+
+* [JENKINS-49596](https://issues.jenkins-ci.org/browse/JENKINS-49596) -
+Prevent User session memory leak by setting the default idle session eviction timeout to 3 minutes.
```
dwnusbaum: Just noticed the default value, does this mean users get logged out of Jenkins if they are inactive for 3 minutes? That seems a little aggressive, from some googling it looks like 20-30 minutes is the usual default for other web servers. (The original reporter of the issue also set it to 30 minutes.)

oleg-nenashev: I am fine with changing default and spinning 4.1.2. @olamy WDYT?

oleg-nenashev: OK, will update

PR #47, opened 2018-03-08
add an option to enable jetty jmx

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
 Changelog
 ===
 
-## 4.2
+## 4.3
 
 Release date: Coming soon
+
+* [PR-47](https://github.com/jenkinsci/winstone/pull/47) - 
+Option to enable [Jetty JMX](https://www.eclipse.org/jetty/documentation/9.4.x/jmx-chapter.html) 
+
+## 4.2
+
+Release date: April 1, 2018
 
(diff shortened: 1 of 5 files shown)


oleg-nenashev:
IIUC it adds [JMX Jetty](https://www.eclipse.org/jetty/documentation/9.4.x/jmx-chapter.html) when CLI option is added. It would be nice for sure.

@olamy  It would be great to have some user-facing documentation in README, e.g.:

* CLI option documentation
* A section, which describes Jetty JMX...
  * Links to official Jetty materials
  * Guides about configuring JMX in Winstone (IIUC it's not enough to use CLI option, Jetty config XMLs need to be updated)

If there are any related tickets in the bugtracker, it would be great to have them as well.

Do you also plan to modify Extras Executable WAR or Jenkins codebases to enable/pre-configure the feature in Jenkins? Or is it a patch for another project using Winstone? Both are fine.

@jenkinsci/code-reviewers FYI

oleg-nenashev:
@olamy Hi, any plans to finalize it soon? If no, I would propose to ship the current master branch state so that the new Jetty version gets adopted in weekly releases early

olamy:
@oleg-nenashev just adding a changelog entry and update documentation here https://wiki.jenkins.io/display/JENKINS/Starting+and+Accessing+Jenkins

PR #48, opened 2018-04-27
typo

docwhat:
I don't know if the same typo exists in the `--help` as well.


oleg-nenashev:
> I don't know if the same typo exists in the --help as well.

As I see, now.
Thanks @docwhat !

PR #54, opened 2018-08-24
use standard jetty qtp (JENKINS-33412, JENKINS-52804, JENKINS-51136)

olamy:


src/main/java/winstone/cmdline/Option.java
```diff
@@ -80,6 +80,10 @@
     public static final OInt CONTROL_PORT=integer("controlPort",-1);
 
     public static final OBoolean USE_JMX=bool("useJmx",false);
+    public static final OBoolean USE_QTP=bool("useQTP",false);
```
olamy: I would definitely prefer having true per default here:
``` public static final OBoolean USE_QTP=bool("useQTP",true); ```

Comment (oleg-nenashev) [reply src/main/java/winstone/cmdline/Option.java]: I am fine with that. CC @jenkinsci/code-reviewers, more votes would be helpful

**Comment by olamy** [comment]
@jenkinsci/code-reviewers

File: README.md [hunk 2]
```diff
@@ -98,7 +95,10 @@ To run different web applications for diffent virtual hosts:
                                   (e.g., xls=application/vnd.ms-excel:wmf=application/x-msmetafile)
        --maxParamCount=N        = set the max number of parameters allowed in a form submission to protect
                                   against hash DoS attack (oCERT #2011-003). Default is 10000.
-       --useJmx                 = Enable Jetty JMX                                   
+       --useJmx                 = Enable Jetty JMX
```
daniel-beck: Seems unused.

oleg-nenashev:
@jenkinsci/code-reviewers I would like to land the change in the next weekly and then to consider it for backporting to .2 or .3. It looks reasonable taking the severity of reported issues.

Any feedback is welcome, esp. from @jenkinsci/core

marco-dev:
Change is tested with a provided WAR on Solaris. Without this change the current Solaris versions of Jenkins are not working. Last working version is LTS 2.107.3. So backporting to last LTS would be great.

olamy:
related issues:
- https://issues.jenkins-ci.org/browse/JENKINS-33412
- https://issues.jenkins-ci.org/browse/JENKINS-52804
- https://issues.jenkins-ci.org/browse/JENKINS-51136

README.md
```diff
@@ -98,7 +95,10 @@ To run different web applications for diffent virtual hosts:
                                   (e.g., xls=application/vnd.ms-excel:wmf=application/x-msmetafile)
        --maxParamCount=N        = set the max number of parameters allowed in a form submission to protect
                                   against hash DoS attack (oCERT #2011-003). Default is 10000.
-       --useJmx                 = Enable Jetty JMX                                   
+       --useJmx                 = Enable Jetty JMX    
+       --qtpMaxThreads          = max threads number when using Jetty Queued Thread Pool
```
oleg-nenashev: Maybe it makes sense to explicitly add "Number" to the argument name. Does not matter much tho

README.md
```diff
@@ -80,12 +80,9 @@ To run different web applications for diffent virtual hosts:
        --httpsKeyManagerType    = the SSL KeyManagerFactory type (eg SunX509, IbmX509). Default is SunX509
        --httpsPrivateKey        = this switch with --httpsCertificate can be used to run HTTPS with OpenSSL secret key
          / --httpsCertificate     file and the corresponding certificate file
-       --http2Port               = set the http2 listening port. -1 to disable, Default is disabled
-       --http2ListenAddress      = set the http2 listening address. Default is all interfaces         
+       --http2Port              = set the http2 listening port. -1 to disable, Default is disabled
+       --http2ListenAddress     = set the http2 listening address. Default is all interfaces         
        --controlPort            = set the shutdown/control port. -1 to disable, Default disabled
-
-       --handlerCountMax        = set the max no of worker threads to allow. Default is 40
```
oleg-nenashev: What will happen if these parameters are defined in startup scripts? IIUC they will be just ignored, right? Even in such case, it makes sense to explicitly document it in the changelog

oleg-nenashev:
Would it justify Winstone 5.0 BTW? Even if we restore binary compatibility, option removal looks important enough to do that.

README.md
svanoort: I think we generally seem to be using "count" - concur that it might make sense to include that, i.e. "qtpMaxThreadCount" and "jettyAcceptorCount" and "jettySelectorCount", etc

But that's more of a style point -- up to the PR author if they want to or not (it's hardly a PR blocker).

src/main/java/winstone/Launcher.java
```diff
@@ -337,9 +305,6 @@ public void shutdown() {
             Logger.log(Logger.INFO, RESOURCES, "Launcher.FailedShutdown", e);
         }
 
-        // Release all listeners/pools/webapps
-        this.threadPool.shutdown();
```
dwnusbaum: Do we still need to clean up the thread pool, or does Jetty handle it now?

src/main/java/winstone/Launcher.java
```diff
@@ -152,8 +145,10 @@ public Launcher(Map args) throws IOException {
                 Thread.currentThread().setContextClassLoader( extraClassLoader );
             }
 
-            this.threadPool = createThreadPool();
-            this.server = new Server(new ExecutorThreadPool(threadPool));
+            int qtpMaxThread = Option.QTP_MAXTHREADS.get(args);
+            QueuedThreadPool queuedThreadPool = qtpMaxThread>0?new QueuedThreadPool(qtpMaxThread):new QueuedThreadPool();
```
dwnusbaum: If the option is not specified, then we use an unbounded thread pool. Do we anticipate any problems with that? We could use a default max value instead to avoid unbounded growth.

skeenan947:
I just tested this on a 72-core amd64 k8s node and it worked perfectly.  Published a docker image built off of this (thanks to @olamy for building the war) at `skeenan947/jenkinstest`

src/main/java/winstone/Launcher.java
rodrigc: @dwnusbaum see results of scalability testing by @skeenan947 here: https://issues.jenkins-ci.org/browse/JENKINS-33412?focusedCommentId=347435&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-347435

dwnusbaum: @rodrigc Is that the right link? I agree that the change is good and works, I just was curious if there is any benefit to using a max limit to the number of threads even in the default case:

    QueuedThreadPool queuedThreadPool = qtpMaxThread > 0
        ? new QueuedThreadPool(qtpMaxThread)
        : new QueuedThreadPool($SOME_DEFAULT_HERE); // Would it be better to put a (large) max here instead of making it unbounded?

olamy: default is 200 https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L72  which seems to be a good value.

README.md
olamy: will generate this:
```
webroot: EnvVars.masterEnvVars.get("JENKINS_HOME")
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at Main._main(Main.java:344)
	at Main.main(Main.java:160)
Caused by: java.lang.IllegalArgumentException: Unrecognized option: --handlerCountMax-=100
	at winstone.cmdline.CmdLineParser.parse(CmdLineParser.java:52)
	at winstone.Launcher.getArgsFromCommandLine(Launcher.java:356)
	at winstone.Launcher.main(Launcher.java:327)
```

olamy: maybe restore them and log a WARNING and sure I will add that to the changelog

olamy:
+1 for 5.0 version I will update that and core pr as well.

src/main/java/winstone/Launcher.java
olamy: Jetty has his own lifecyle see doSop here https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L150

src/main/java/winstone/Launcher.java
dwnusbaum: Thanks @olamy looks like I was just confused!

README.md
olamy: @svanoort I like this naming proposal! Style it's very important! :)

PR #57, opened 2018-11-08
Working around SUREFIRE-1588

jglick:
Picking up https://github.com/jenkinsci/pom/pull/34. Maybe an uncomplicated example at last? Really all of these are snowflakes.


jglick:
@batm

jglick:
or @jvz or @olamy

PR #60, opened 2019-03-18
JENKINS-56591 cipher exclusion configurable

olamy:
since jetty 9.4.12.v20180830 default cipher exclusion has been changed with https://github.com/eclipse/jetty.project/issues/2807. We need  a configuration parameter so the user can choose their prefer exclusion list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,10 @@ To run different web applications for diffent virtual hosts:
 
 
 ## Command-line options
-    Usage: java winstone.Launcher [--option=value] [--option=value] [etc]
 
+    Winstone Servlet Engine v4.0, (c) 2003-2006 Rick Knowles
+    Usage: java winstone.jar [--option=value] [--option=value] [etc]
+    
     Required options: either --webroot OR --warfile OR --webappsDir OR --hostsDir
        --webroot                = set document root folder.
        --warfile                = set location of warfile to extract from.
@@ -63,12 +65,13 @@ To run different web applications for diffent virtual hosts:
        --config                 = load configuration properties from here. Default is ./winstone.properties
        --prefix                 = add this prefix to all URLs (eg http://localhost:8080/prefix/resource). Default is none
        --commonLibFolder        = folder for additional jar files. Default is ./lib
+    
        --extraLibFolder         = folder for additional jar files to add to Jetty classloader
-
+    
        --logThrowingLineNo      = show the line no that logged the message (slow). Default is false
        --logThrowingThread      = show the thread that logged the message. Default is false
        --debug                  = set the level of debug msgs (1-9). Default is 5 (INFO level)
-
+    
        --httpPort               = set the http listening port. -1 to disable, Default is 8080
        --httpListenAddress      = set the http listening address. Default is all interfaces
        --httpKeepAliveTimeout   = how long idle HTTP keep-alive connections are kept around (in ms; default 5000)?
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -51,24 +51,28 @@ To run different web applications for diffent virtual hosts:
...
+    Usage: java -jar jenkins.war [--option=value] [--option=value]
+    
+    Options:
+       --webroot                = folder where the WAR file is expanded into. Default is ${JENKINS_HOME}/war
+       --pluginroot             = folder where the plugin archives are expanded into. Default is ${JENKINS_HOME}/plugins
+                                  (NOTE: this option does not change the directory where the plugin archives are stored)
+       --extractedFilesFolder   = folder where extracted files are to be located. Default is the temp folder
+       --daemon                 = fork into background and run as daemon (Unix only)
+       --logfile                = redirect log messages to this file
+       --enable-future-java     = allows running with new Java versions which are not fully supported (class version 52 and above)
```
Wadeck: When I look at the other options, it seems the "implicit convention" was to use camelCase instead of kebab-case. Not sure it's really related to this PR as I do not see the change of that option.

src/main/java/winstone/AbstractSecuredConnectorFactory.java
```diff
@@ -197,14 +199,20 @@ protected SslContextFactory getSSLContext( Map args) {
...
-
             ssl.setKeyStore(keystore);
             ssl.setKeyStorePassword(keystorePassword);
             ssl.setKeyManagerPassword(privateKeyPassword);
             ssl.setKeyManagerFactoryAlgorithm(Option.HTTPS_KEY_MANAGER_TYPE.get(args));
             ssl.setCertAlias(Option.HTTPS_CERTIFICATE_ALIAS.get(args));
             ssl.setExcludeProtocols("SSLv3", "SSLv2", "SSLv2Hello");
-
+            String excludeCiphers = Option.HTTPS_EXCLUDE_CIPHER_SUITES.get( args );
+            if( StringUtil.isNotBlank(excludeCiphers)) {
```
Wadeck: I think it worth mentioning in the description of the option how to provide "no exclusion". Passing an empty string will let the default exclusion in place.

README.md
oleg-nenashev: I have no idea why it has to be in Winstone at all. Should be moved to https://github.com/jenkinsci/extras-executable-war IMHO

src/main/java/winstone/AbstractSecuredConnectorFactory.java
```diff
@@ -197,14 +199,20 @@ protected SslContextFactory getSSLContext( Map args) {
...
+                String[] cipherSuites = excludeCiphers.split(",");
```
oleg-nenashev:
```suggestion
                String[] cipherSuites = excludeCiphers.split("\\s*,\\s*");
```

oleg-nenashev: Note that it will not be handled properly if the user-specified regexp includes `,` due to whatever reason

README.md
olamy: oops my bad I copied the output of jenkins.war :)

olamy: fixed

src/main/java/winstone/AbstractSecuredConnectorFactory.java
olamy: ok (well not sure to have cipher name including `,` :) but it's a very simple change

src/main/java/winstone/AbstractSecuredConnectorFactory.java
olamy: good point!
the no exclusion is currently a fake exclusion such ` --excludeCipherSuites=BEER` 
maybe we can let users pass an empty string if they want no exclusion? (using isNotEmpty rather than isNotBlank)
WDYT?

Wadeck: I would prefer such approach than the random string :) Imagine the admin that should maintain such configuration, good luck to find why the previous admin put a random string (that is not random for him)

src/main/java/winstone/AbstractSecuredConnectorFactory.java
Wadeck: Correct me if I am wrong, but if I pass: `RSA_X,RSA_Y` (2x), it will pass as expected (before and after). But if I pass `RSA_X,Y_OR_Z` (1x), the behavior is exactly the same, even for `RSA_X, Y_OR_Z` (1x).

The only thing that is changed is in case I pass `RSA_X, RSA_Y` (2x) with the space after the comma, before the patch, the second regex contains the comma and after the correction it will not. So the correction is about user-specified regexp separated by `,` with spaces, not directly in the regexp right?

Please propose samples that show the faulty behavior if my hypothesis is not correct.

olamy: @Wadeck you're correct I'm not sure how to handle `,` but I'm not sure if we have to manage this as I cannot really find any cipher with `,` in his name... See https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4
So I would simply use the first version with `  String[] cipherSuites = excludeCiphers.split(","); `
and voila no real need here to make this more complicated...
?

src/main/java/winstone/AbstractSecuredConnectorFactory.java
olamy: done possible to exclude NO cipher by using ` --excludeCipherSuites=' ' `

basil:
Caused #437.

PR #62, opened 2019-04-03
Stream does not need to be closed given try-with-resources is in place.

varyvol:
try-with-resources was introduced in [this commit](https://github.com/jenkinsci/winstone/commit/aed91a80cd509da955f749c4539bc3599ef87ee6#commitcomment-33011448) and therefore the stream must not be closed manually.

@olamy


olamy:
@varyvol thanks!

PR #69, opened 2019-08-20
Switch to spotbugs and fix some issues

res0nance:
- Swapped to spotbugs and fixed multiple issues.
- There still remains some issues but the PR was getting large

### DIFF
--- a/src/main/java/winstone/Ajp13ConnectorFactory.java
+++ b/src/main/java/winstone/Ajp13ConnectorFactory.java
@@ -22,7 +22,6 @@ import java.util.Map;
 public class Ajp13ConnectorFactory implements ConnectorFactory {
     public boolean start(Map args, Server server) throws IOException {
         int listenPort = Option.AJP13_PORT.get(args);
-        // String listenAddress = Option.AJP13_LISTEN_ADDRESS.get(args);
 
         if (listenPort < 0) {
             return false;
--- a/src/main/java/winstone/Ajp13ConnectorFactory.java
+++ b/src/main/java/winstone/Ajp13ConnectorFactory.java
@@ -15,14 +15,14 @@ import java.util.Map;
 /**
  * Implements the main listener daemon thread. This is the class that gets
  * launched by the command line, and owns the server socket, etc.
- * 
+ *
  * @author mailto: <a href="rick_knowles@hotmail.com">Rick Knowles</a>
  * @version $Id: Ajp13ConnectorFactory.java,v 1.12 2006/03/24 17:24:22 rickknowles Exp $
  */
 public class Ajp13ConnectorFactory implements ConnectorFactory {
     public boolean start(Map args, Server server) throws IOException {
         int listenPort = Option.AJP13_PORT.get(args);
-        String listenAddress = Option.AJP13_LISTEN_ADDRESS.get(args);
+        // String listenAddress = Option.AJP13_LISTEN_ADDRESS.get(args);
 
         if (listenPort < 0) {
             return false;
--- a/src/main/java/winstone/HostConfiguration.java
+++ b/src/main/java/winstone/HostConfiguration.java
@@ -24,6 +24,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Constructor;
 import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Enumeration;
@@ -235,8 +236,11 @@ public class HostConfiguration {
                         (userName != null ? WinstoneResourceBundle.globalReplace(userName,
                                 new String[][] {{"/", ""}, {"\\", ""}, {",", ""}}) + "/" : "") +
                         "winstone/" + warfile.getName());
-                if (!tempFile.delete()) {
-                    Logger.logDirectMessage(Logger.WARNING, null, "Failed To delete dummy file", null);
+
+                try {
+                    Files.delete(tempFile.toPath());
+                } catch (Exception ex) {
+                    Logger.logDirectMessage(Logger.WARNING, null, "Failed To delete dummy file", ex);
                 }
             }
             if (unzippedDir.exists()) {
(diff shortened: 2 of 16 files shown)


oleg-nenashev:
Need to review it carefully, but thanks a lot!

oleg-nenashev:
sorry, still in my queue

src/main/java/winstone/Ajp13ConnectorFactory.java
```diff
@@ -15,14 +15,14 @@
...
- * 
+ *
  * @author mailto: <a href="rick_knowles@hotmail.com">Rick Knowles</a>
  * @version $Id: Ajp13ConnectorFactory.java,v 1.12 2006/03/24 17:24:22 rickknowles Exp $
  */
 public class Ajp13ConnectorFactory implements ConnectorFactory {
     public boolean start(Map args, Server server) throws IOException {
         int listenPort = Option.AJP13_PORT.get(args);
-        String listenAddress = Option.AJP13_LISTEN_ADDRESS.get(args);
+        // String listenAddress = Option.AJP13_LISTEN_ADDRESS.get(args);
```
oleg-nenashev: If unused, please.do one of the following..

* delete it
* Create a follow-up bug, link it from a TODO comment

Looks like it is a bug unless I miss something

src/main/java/winstone/HostConfiguration.java
```diff
@@ -250,43 +252,51 @@ protected File getWebRoot(File requestedWebroot, File warfile) throws IOExceptio
...
-                    while ( readBytes != -1 ) {
-                        outStream.write( buffer, 0, readBytes );
-                        readBytes = inContent.read( buffer );
+                    // If archive date is newer than unzipped file, overwrite
+                    File outFile = new File(unzippedDir, elemName);
+                    if (outFile.exists() && (outFile.lastModified() > warfile.lastModified())) {
+                        continue;
+                    }
+                    File parentOutFile = outFile.getParentFile();
+                    if (!parentOutFile.mkdirs()) {
```
oleg-nenashev: We can drop Java 6 later and start using `Files` here.
CC @olamy

src/main/java/winstone/realm/FileRealm.java
```diff
@@ -34,12 +34,12 @@
...
 public class FileRealm extends HashLoginService {
     private static final WinstoneResourceBundle REALM_RESOURCES = new WinstoneResourceBundle("winstone.realm.LocalStrings");
-    
-    final String DEFAULT_FILE_NAME = "users.xml";
-    final String ELEM_USER = "user";
-    final String ATT_USERNAME = "username";
-    final String ATT_PASSWORD = "password";
-    final String ATT_ROLELIST = "roles";
+
+    static final String DEFAULT_FILE_NAME = "users.xml";
```
oleg-nenashev: Breaks public API?

res0nance: I don;t see how you should still be able to access the variable via a static or nonstatic way

src/main/java/winstone/HostConfiguration.java
res0nance: I'll clean this bit up then

src/main/java/winstone/realm/FileRealm.java
oleg-nenashev: Right, sorry

src/main/java/winstone/tools/WinstoneControl.java
```diff
@@ -22,20 +22,20 @@
...
-    
+
     final static String OPERATION_SHUTDOWN = "shutdown";
     final static String OPERATION_RELOAD = "reload:";
     static int TIMEOUT = 10000;
 
-    public static OInt CONTROL_PORT = Option.integer("controlPort");
-    public static OInt PORT = Option.integer("port");
-    public static OInt DEBUG = new OInt("debug", 5) {
+    public final static OInt CONTROL_PORT = Option.integer("controlPort");
```
oleg-nenashev: IIRC breaking change was here. Not that important

src/main/java/winstone/HostConfiguration.java
```diff
@@ -230,10 +232,12 @@ protected File getWebRoot(File requestedWebroot, File warfile) throws IOExceptio
                 File tempFile = File.createTempFile("dummy", "dummy");
                 String userName = System.getProperty("user.name");
                 unzippedDir = new File(tempFile.getParent(),
-                        (userName != null ? WinstoneResourceBundle.globalReplace(userName, 
+                        (userName != null ? WinstoneResourceBundle.globalReplace(userName,
                                 new String[][] {{"/", ""}, {"\\", ""}, {",", ""}}) + "/" : "") +
                         "winstone/" + warfile.getName());
-                tempFile.delete();
+                if (!tempFile.delete()) {
```
jglick: Or there is a Commons IO utility, or you can use `java.nio.file.Files.delete` which throws a meaningful exception.

src/main/java/winstone/HostConfiguration.java
```diff
@@ -250,43 +254,52 @@ protected File getWebRoot(File requestedWebroot, File warfile) throws IOExceptio
             if(!timestampFile.exists() || Math.abs(timestampFile.lastModified()- warfile.lastModified())>1000) {
                 // contents of the target directory is inconsistent from the war.
                 deleteRecursive(unzippedDir);
-                unzippedDir.mkdirs();
+                if (!unzippedDir.mkdirs()) {
```
jglick: ditto

src/main/java/winstone/HostConfiguration.java
```diff
@@ -250,43 +254,52 @@ protected File getWebRoot(File requestedWebroot, File warfile) throws IOExceptio
...
-                    while ( readBytes != -1 ) {
-                        outStream.write( buffer, 0, readBytes );
-                        readBytes = inContent.read( buffer );
+                    // If archive date is newer than unzipped file, overwrite
+                    File outFile = new File(unzippedDir, elemName);
+                    if (outFile.exists() && (outFile.lastModified() > warfile.lastModified())) {
+                        continue;
+                    }
+                    try {
+                        Files.createDirectories(Paths.get(outFile.getParent()));
```
jglick: FWIW in Jenkins core we have been catching `InvalidPathException` as well, which apparently can happen under some obscure conditions on Windows servers.

oleg-nenashev: Would recommend to address it before merging

PR #79, opened 2019-11-22
[JEP-222] - Add WebSocket support

jglick:
See https://github.com/jenkinsci/jep/pull/250.


oleg-nenashev:
I will merge it towards the next version.
If there is serious performance degradation or any regression, we can always revert it.
Still, I would expect some performance evaluation in the JEP

jglick:
It should have no effect if core support for WebSocket is unmerged. Just a slightly larger `winstone.jar`.

PR #86, opened 2020-01-03
Various regressions from earlier code cleanup

jglick:
Corrects a regression from #69 found by @jeffret-b: https://github.com/jenkinsci/jenkins/pull/4369#issuecomment-570639297

See: https://github.com/jenkinsci/extras-executable-war/blob/2af94677c0ba557351d2c0703c3c43030c198b50/src/main/java/Main.java#L343-L371

### DIFF
--- a/src/main/java/winstone/WinstoneSession.java
+++ b/src/main/java/winstone/WinstoneSession.java
@@ -9,5 +9,9 @@ public class WinstoneSession {
     /**
      * Name of the cookie that stores HTTP session ID.
      */
-    public final static String SESSION_COOKIE_NAME = "JSESSIONID";
+    // @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "mutated by extras-executable-war")
+    public static String SESSION_COOKIE_NAME = "JSESSIONID";
+
+    private WinstoneSession() {}
+
 }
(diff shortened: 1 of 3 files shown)


src/main/java/winstone/WinstoneSession.java
```diff
@@ -9,5 +9,9 @@
     /**
      * Name of the cookie that stores HTTP session ID.
      */
-    public final static String SESSION_COOKIE_NAME = "JSESSIONID";
+    // @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "mutated by extras-executable-war")
```
jglick: #69 did not add the annotations JAR as a dep and I did not care to deal with that here. Anyway SB errors are not currently fatal.

jglick:
By the way @oleg-nenashev [ATH caught this](https://ci.jenkins.io/job/Core/job/jenkins/job/PR-4369/34/testReport/junit/core/FreestyleJobTest/ath___Running_ATH___ATH_categories_firefox_jdk8___buildParametrized/) after integration into Jenkins.

oleg-nenashev:
Ack, thanks for fixing. It would be great to enable the integration testing everywhere :(

res0nance:
Thanks for fixing the issues, also shoutouts to quality testing in ath

PR #91, opened 2020-01-24
[JENKINS-60857] Using SslContextFactory.Server

jglick:
Guessing this would solve [JENKINS-60857](https://issues.jenkins-ci.org/browse/JENKINS-60857), but I have yet to reproduce that.

The bug seems to have been introduced by https://github.com/eclipse/jetty.project/pull/4326, which was subsequently amended at least in https://github.com/eclipse/jetty.project/pull/4386 & https://github.com/eclipse/jetty.project/pull/4404.


oleg-nenashev:
@jglick Should I ship it?

jglick:
If you are awake, sure! Otherwise I will get to it tonight or tomorrow I guess.

PR #125, opened 2020-10-19
downgrade jetty to 9.4.30.v20200611

olamy:


oleg-nenashev (approved):
CI issue is a known failure

pom.xml
```diff
@@ -17,9 +17,9 @@
   <url>https://github.com/jenkinsci/winstone</url>
 
   <properties>
-    <revision>5.12</revision>
+    <revision>5.11.1</revision>
```
jglick, line 20: Do not do this sort of thing in `master`. Use e.g. https://gist.github.com/jglick/86a30894446ed38f918050c1180483e2

pom.xml
```diff
@@ -17,9 +17,9 @@
...
     <changelist>-SNAPSHOT</changelist>
-    <jetty.version>9.4.32.v20200930</jetty.version>
+    <jetty.version>9.4.30.v20200611</jetty.version>
```
jglick, line 22: I thought we were at 9.4.31? Or was this the last version used in 5.11?

olamy: 5.10 was 9.4.30.v20200611 see https://github.com/jenkinsci/winstone/blob/2bd71ffedba612833704d734e1323223e52a23d7/pom.xml#L22

PR #126, opened 2020-10-20
`--requestHeaderSize` argument is added to the help string.

duemir:


duemir:
Doesn't look like the check failure has anything to do with what I have done. The master commits seem to be failing with similar.

olamy:
@duemir yup windows is a pain :)

PR #165, opened 2021-08-16
[JENKINS-66379] use default keystore format

jtnord:
[JENKINS-66379](https://issues.jenkins.io/browse/JENKINS-66379)
The JKS keystore is insecure and is some environments not even allowed.

We do not care what format the keystore is in, so just use the default.

If someone has a non default keystore yet is using a keystore here then
this will break on upgrade, however it is unlikely that they would be
having a non default keystore as the normal reason for doing that is due
to a restriction like FIPS-140 and given the keystore we require has a
key using JDK here would be non compliant.

no unit tests as this requires changing java security properties for the JVM.  Existing tests will ensure there are no regressions for the normal cases.


timja (approved):
ref https://docs.oracle.com/javase/8/docs/api/java/security/KeyStore.html this is now using the recommended APIs

jvz (approved):
Thanks for taking care of this!

oleg-nenashev (approved):
I approve that, but IMO we should document it as a potentially breaking change in the Jenkins core and upgrade guide. Is it correct @jtnord ?

olamy (approved):

PR #166, opened 2021-08-16
reduce test flakyness

jtnord:
ise a port from the user range rather than the dynamic range to reduce
the chance of collision.
Additionally it is just the SSL port that needs to be known so using a
random free port for the http socket further reduces the chances of a
port being in use

- [x] Link to relevant pull requests, esp. upstream and downstream changes


oleg-nenashev (approved):
+1 from me. Bonus point for automatic collision detector

olamy:
ah yup good one. I saw my TODO :)
So I would even prefer everything with dynamic port see https://github.com/jenkinsci/winstone/pull/167

olamy:
replaced by https://github.com/jenkinsci/winstone/pull/167

PR #181, opened 2021-12-19
Upgrade Jetty to 10.0.11

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
     ignore:
       # Restrict updates to jetty in the 9.4.x space
       - dependency-name: "org.eclipse.jetty:*"
-        versions: [ ">=10" ]
+        versions: [ ">=11" ]
 
   - package-ecosystem: "github-actions"
     directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
       interval: "weekly"
     target-branch: "master"
     ignore:
-      # Restrict updates to jetty in the 9.4.x space
+      # Restrict updates to jetty in the 10.x space
       - dependency-name: "org.eclipse.jetty:*"
         versions: [ ">=11" ]
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,5 @@
  */
 buildPlugin(useContainerAgent: true, configurations: [
   [ platform: 'linux', jdk: '11' ],
-  [ platform: 'linux', jdk: '17' ],
   [ platform: 'windows', jdk: '11' ]
 ])
-
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@
  */
 buildPlugin(useContainerAgent: true, configurations: [
   [ platform: 'linux', jdk: '11' ],
+  [ platform: 'linux', jdk: '17' ],
   [ platform: 'windows', jdk: '11' ]
 ])
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,3 +7,4 @@ buildPlugin(useContainerAgent: true, configurations: [
   [ platform: 'linux', jdk: '11' ],
   [ platform: 'windows', jdk: '11' ]
 ])
+
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ To run different web applications for diffent virtual hosts:
          / --httpsCertificate     file and the corresponding certificate file
        --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)
        --http2Port              = set the http2 listening port. -1 to disable, Default is disabled
+       -- httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
+       --httpsSniRequired       = if a SNI certificate is required. Disabled per default
        --http2ListenAddress     = set the http2 listening address. Default is all interfaces
        --excludeCipherSuites    = set the ciphers to exclude (comma separated, use blank quote " " to exclude none) (default is 
                                // Exclude weak / insecure ciphers 
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ To run different web applications for diffent virtual hosts:
          / --httpsCertificate     file and the corresponding certificate file
        --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)
        --http2Port              = set the http2 listening port. -1 to disable, Default is disabled
-       --httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
+       --httpsSniHostCheck      = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
        --httpsSniRequired       = if a SNI certificate is required. Disabled per default
        --http2ListenAddress     = set the http2 listening address. Default is all interfaces
        --excludeCipherSuites    = set the ciphers to exclude (comma separated, use blank quote " " to exclude none) (default is 
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ To run different web applications for diffent virtual hosts:
          / --httpsCertificate     file and the corresponding certificate file
        --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)
        --http2Port              = set the http2 listening port. -1 to disable, Default is disabled
-       -- httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
+       --httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
        --httpsSniRequired       = if a SNI certificate is required. Disabled per default
        --http2ListenAddress     = set the http2 listening address. Default is all interfaces
        --excludeCipherSuites    = set the ciphers to exclude (comma separated, use blank quote " " to exclude none) (default is 
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
   <properties>
     <revision>6.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jetty.version>10.0.7</jetty.version>
+    <jetty.version>10.0.9</jetty.version>
     <alpn.api.version>1.1.3.v20160715</alpn.api.version>
     <!-- TODO: remove once SpotBugs issues are fixed. 29 so far -->
     <spotbugs.threshold>High</spotbugs.threshold>
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,6 @@
     <changelist>-SNAPSHOT</changelist>
     <jetty.version>10.0.7</jetty.version>
     <alpn.api.version>1.1.3.v20160715</alpn.api.version>
-    <java.target>11</java.target>
     <!-- TODO: remove once SpotBugs issues are fixed. 29 so far -->
     <spotbugs.threshold>High</spotbugs.threshold>
     <gitHubRepo>jenkinsci/winstone</gitHubRepo>
@@ -347,7 +346,7 @@
         <jdk>17</jdk>
       </activation>
       <properties>
-        <argLine>--add-opens=java.base/sun.security.util=ALL-UNNAMED --add-opens=java.base/sun.security.tools.keytool=ALL-UNNAMED --add-opens=java.base/sun.security.x509=ALL-UNNAMED</argLine>
+        <argLine>--add-opens=java.base/sun.security.util=ALL-UNNAMED</argLine>
       </properties>
     </profile>
   </profiles>
--- a/pom.xml
+++ b/pom.xml
@@ -19,8 +19,9 @@
   <properties>
     <revision>6.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jetty.version>9.4.48.v20220622</jetty.version>
+    <jetty.version>10.0.7</jetty.version>
     <alpn.api.version>1.1.3.v20160715</alpn.api.version>
+    <java.target>11</java.target>
     <!-- TODO: remove once SpotBugs issues are fixed. 29 so far -->
     <spotbugs.threshold>High</spotbugs.threshold>
     <gitHubRepo>jenkinsci/winstone</gitHubRepo>
@@ -159,6 +160,24 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>display-info</id>
+            <configuration>
+              <rules>
+                <enforceBytecodeVersion>
+                  <excludes combine.children="append">
+                    <exclude>org.eclipse.jetty*:*</exclude>
+                  </excludes>
+                </enforceBytecodeVersion>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 
@@ -284,8 +303,13 @@
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty.websocket</groupId>
-      <artifactId>websocket-server</artifactId> <!-- or javax-websocket-server-impl -->
+      <artifactId>websocket-jetty-server</artifactId> <!-- or websocket-javax-server -->
       <version>${jetty.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>2.0.0-alpha5</version>
     </dependency>
 
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
   <properties>
     <revision>6.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jetty.version>10.0.9</jetty.version>
+    <jetty.version>10.0.11</jetty.version>
     <alpn.api.version>1.1.3.v20160715</alpn.api.version>
     <!-- TODO: remove once SpotBugs issues are fixed. 29 so far -->
     <spotbugs.threshold>High</spotbugs.threshold>
--- a/pom.xml
+++ b/pom.xml
@@ -288,6 +288,12 @@
     <dependency>
       <groupId>org.eclipse.jetty.websocket</groupId>
       <artifactId>websocket-jetty-server</artifactId> <!-- or websocket-javax-server -->
+      <exclusions>
+        <exclusion>
+          <groupId>org.eclipse.jetty</groupId>
+          <artifactId>jetty-annotations</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+<!--          Uncomment to enable extra logging for debugging          -->
 <!--          <systemPropertyVariables>-->
 <!--            <java.util.logging.config.file>src/test/resources/jul.properties</java.util.logging.config.file>-->
 <!--          </systemPropertyVariables>-->
--- a/pom.xml
+++ b/pom.xml
@@ -327,16 +327,4 @@
     </dependency>
   </dependencies>
 
-  <profiles>
-    <profile>
-      <id>jdk17</id>
-      <activation>
-        <jdk>17</jdk>
-      </activation>
-      <properties>
-        <argLine>--add-opens=java.base/sun.security.util=ALL-UNNAMED</argLine>
-      </properties>
-    </profile>
-  </profiles>
-
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -73,6 +73,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+<!--          <systemPropertyVariables>-->
+<!--            <java.util.logging.config.file>src/test/resources/jul.properties</java.util.logging.config.file>-->
+<!--          </systemPropertyVariables>-->
           <trimStackTrace>false</trimStackTrace> <!-- SUREFIRE-1226 workaround -->
         </configuration>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -340,4 +340,16 @@
     </dependency>
   </dependencies>
 
+  <profiles>
+    <profile>
+      <id>jdk17</id>
+      <activation>
+        <jdk>17</jdk>
+      </activation>
+      <properties>
+        <argLine>--add-opens=java.base/sun.security.util=ALL-UNNAMED --add-opens=java.base/sun.security.tools.keytool=ALL-UNNAMED --add-opens=java.base/sun.security.x509=ALL-UNNAMED</argLine>
+      </properties>
+    </profile>
+  </profiles>
+
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -223,6 +223,13 @@
         <version>1.0</version>
         <type>signature</type>
       </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-bom</artifactId>
+        <version>${jetty.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
@@ -262,37 +269,30 @@
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-server</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-util</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-servlet</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-webapp</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty.http2</groupId>
       <artifactId>http2-server</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty.http2</groupId>
       <artifactId>http2-hpack</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-alpn-server</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty.alpn</groupId>
@@ -302,12 +302,10 @@
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-jmx</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty.websocket</groupId>
       <artifactId>websocket-jetty-server</artifactId> <!-- or websocket-javax-server -->
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -299,7 +299,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-jdk14</artifactId>
-      <version>2.0.0-alpha5</version>
+      <version>2.0.0-alpha6</version>
     </dependency>
 
     <dependency>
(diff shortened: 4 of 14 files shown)


pom.xml
```diff
@@ -19,9 +19,9 @@
   <properties>
     <revision>5.22</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jetty.version>9.4.44.v20210927</jetty.version>
+    <jetty.version>10.0.7</jetty.version>
     <alpn.api.version>1.1.3.v20160715</alpn.api.version>
-    <java.level>8</java.level>
+    <java.target>11</java.target>
```
basil, line 24: This change has no effect (see for yourself by setting `<java.target>nonsense</java.target>`. I suggest reverting.

pom.xml
```diff
@@ -284,9 +302,14 @@
...
       <groupId>org.eclipse.jetty.websocket</groupId>
-      <artifactId>websocket-server</artifactId> <!-- or javax-websocket-server-impl -->
+      <artifactId>websocket-jetty-server</artifactId> <!-- or websocket-javax-server -->
       <version>${jetty.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>2.0.0-alpha5</version>
+    </dependency>
```
basil, line 312: Not sure I see the reason for adding this dependency. Seems to work fine without this change for me.

src/main/java/winstone/AbstractSecuredConnectorFactory.java
```diff
@@ -133,8 +134,9 @@ private static PrivateKey readPEMRSAPrivateKey(Reader reader) throws IOException
                         in = !in;
                         continue;
                     }
+
                     if ( in ) {
-                        baos.write(Base64.getDecoder().decode(line.trim()));
+                        baos.write( Base64.getDecoder().decode(line.getBytes(StandardCharsets.UTF_8) ) );
```
basil, line 139: Seems unrelated to the basic point of this change. Better suited for a separate PR I think.

basil (commented):
A promising start! Needs `Jenkinsfile` changes for a Java 11 runtime, and these three tests are failing:

- `HttpsConnectorFactoryTest#httpRedirect`
- `HttpsConnectorFactoryTest#testHttpsRandomCert`
- `HttpsConnectorFactoryTest#wildcard`

src/main/java/winstone/AbstractSecuredConnectorFactory.java
olamy: because I started this long time ago. and this jetty class has been removed in jetty 10 so I had to change it when starting this branch

pom.xml
olamy: do you have logs without this?

olamy: because Jetty 10 is based only on slf4j so need a slf4j implementation available in the classpath

pom.xml
olamy: yup the parent pom and format of this is a real pain... I have started some changed in a local parent pom (but I havent commit it)

olamy:
> A promising start! Needs `Jenkinsfile` changes for a Java 11 runtime, and these three tests are failing:
> 
> * `HttpsConnectorFactoryTest#httpRedirect`
> * `HttpsConnectorFactoryTest#testHttpsRandomCert`
> * `HttpsConnectorFactoryTest#wildcard`

I know and that's why it's still draft. and btw I cannot push this in master as it's java11 mandatory

pom.xml
basil: Good point. Without this I get:

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#noProviders for further details.
```

So I guess this is necessary.

README.md
```diff
@@ -85,6 +85,8 @@ To run different web applications for diffent virtual hosts:
          / --httpsCertificate     file and the corresponding certificate file
        --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)
        --http2Port              = set the http2 listening port. -1 to disable, Default is disabled
+       -- httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
```
jtnord, line 88:
```suggestion
       --httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
```

src/main/resources/winstone/LocalStrings.properties
```diff
@@ -98,6 +98,8 @@ Launcher.UsageInstructions.Options=\
 \     / --httpsCertificate     file and the corresponding certificate file\n\
 \   --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)\n\
 \   --http2Port              = set the http2 listening port. -1 to disable, Default is disabled\n\
+\   -- httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default\n\
```
jtnord, line 101:
```suggestion
\   --httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default\n\
```

jtnord:
fails in java17 due to
```
Caused by: winstone.WinstoneException: Failed to create a self-signed certificate; make one yourself.
	at winstone.AbstractSecuredConnectorFactory.configureSsl(AbstractSecuredConnectorFactory.java:111)
	at winstone.HttpsConnectorFactory.start(HttpsConnectorFactory.java:54)
	at winstone.Launcher.spawnListener(Launcher.java:219)
	... 31 more
Caused by: java.lang.IllegalAccessException: class winstone.AbstractSecuredConnectorFactory cannot access class sun.security.tools.keytool.CertAndKeyGen (in module java.base) because module java.base does not export sun.security.tools.keytool to unnamed module @5274766b
```

olamy:
> fails in java17 due to
> 
> ```
> 
> Caused by: winstone.WinstoneException: Failed to create a self-signed certificate; make one yourself.
> 
> 	at winstone.AbstractSecuredConnectorFactory.configureSsl(AbstractSecuredConnectorFactory.java:111)
> 
> 	at winstone.HttpsConnectorFactory.start(HttpsConnectorFactory.java:54)
> 
> 	at winstone.Launcher.spawnListener(Launcher.java:219)
> 
> 	... 31 more
> 
> Caused by: java.lang.IllegalAccessException: class winstone.AbstractSecuredConnectorFactory cannot access class sun.security.tools.keytool.CertAndKeyGen (in module java.base) because module java.base does not export sun.security.tools.keytool to unnamed module @5274766b
> 
> ```

Yup ezy fix

olamy:
@jtnord java17 build fix but it requires some special jvm/jpms flags. not sure this will be something user friendly.
choices:
- remove this `make home made certificate on the fly` feature (not sure how. it is used)
- use something such bouncycastle to do it (not too much complicated)

not sure as we are not yet to be java 17 mandatory 🤣.

wdyt?

timja:
It’s no issue to add the flags for Java17 but we probably shouldn’t be using the key tool module, if the JDK doesn’t provide the required APIs then we should take a look at bouncy castle which normally does :)

jtnord:
> It’s no issue to add the flags for Java17 but we probably shouldn’t be using the key tool module, if the JDK doesn’t provide the required APIs then we should take a look at bouncy castle which normally does :)

you really should not use bouncycastle inside core or any library it uses. bouncycastle is a plugin and having it in core will create issues, and make it really hard for people wanting to run Jenkins in a FIPS environment.  A lot of work was done to split BC out of core previously. [JENKINS-35291](https://issues.jenkins.io/browse/JENKINS-35291) and related tickets

timja:
I would just remove the feature, it's using sha1 certs which don't work anymore in browsers as far as I know

Jenkinsfile
```diff
@@ -4,7 +4,8 @@
  * allowing one to test against multiple Jenkins versions.
  */
 buildPlugin(useContainerAgent: true, configurations: [
-  [ platform: 'linux', jdk: '8' ],
   [ platform: 'linux', jdk: '11' ],
+  [ platform: 'linux', jdk: '17' ],
```
timja, line 8: doesn't appear to be working:

```
[](https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fwinstone/detail/PR-181/15/pipeline/48#step-55-log-2)[](https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fwinstone/detail/PR-181/15/pipeline/48#step-55-log-3)[](https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fwinstone/detail/PR-181/15/pipeline/48#step-55-log-4)[](https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fwinstone/detail/PR-181/15/pipeline/48#step-55-log-5)[](https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fwinstone/detail/PR-181/15/pipeline/48#step-55-log-6)[2022-02-08T03:41:05.381Z] + mvn --batch-mode --show-version --errors --no-transfer-progress --update-snapshots -Dmaven.repo.local=/home/jenkins/workspace/Core_winstone_PR-181@tmp/m2repo -Dmaven.test.failure.ignore -Dspotbugs.failOnError=false -Dcheckstyle.failOnViolation=false -Dcheckstyle.failsOnError=false -Penable-jacoco -Dset.changelist help:evaluate -Dexpression=changelist -Doutput=/home/jenkins/workspace/Core_winstone_PR-181@tmp/changelist clean install

[2022-02-08T03:41:05.381Z] Apache Maven 3.8.4 (9b656c72d54e5bacbed989b64718c159fe39b537)

[2022-02-08T03:41:05.381Z] Maven home: /usr/share/maven

[2022-02-08T03:41:05.381Z] Java version: 11.0.13, vendor: Eclipse Adoptium, runtime: /opt/java/openjdk

[2022-02-08T03:41:05.381Z] Default locale: en_US, platform encoding: UTF-8

[2022-02-08T03:41:05.381Z] OS name: "linux", version: "5.4.172-90.336.amzn2.x86_64", arch: "amd64", family: "unix"
```

olamy: hehe good catch I was blindly trusted the green

src/main/java/winstone/Launcher.java
```diff
@@ -347,11 +346,10 @@ public static void main(String[] argv) throws IOException {
         if (System.getProperty("java.util.logging.config.file") == null) {
           for (Handler h : java.util.logging.Logger.getLogger("").getHandlers()) {
               if (h instanceof ConsoleHandler) {
-                  ((ConsoleHandler) h).setFormatter(new SupportLogFormatter());
+                  h.setFormatter(new SupportLogFormatter());
```
basil, line 349: I do not see how this change is related to the Jetty 10 upgrade. Suggest leaving unrelated changes to separate PRs for readability.

basil (changes requested):

src/main/java/winstone/Launcher.java
olamy: you are right my bad. This cosmetic change in a middle of a documented breaking change definitely need his own PR, extended review and deserves his own entry in the release notes
here you go https://github.com/jenkinsci/winstone/pull/207

pom.xml
jtnord: but version 2?  why is jetty using an [unstable / experimental](https://www.slf4j.org/download.html#content) version.

will this impact jenkins and all plugins?  Many plugins are using slf4j APIs - do we have a another potential guavopolitc issue here?

joakime: > but version 2? why is jetty using an [unstable / experimental](https://www.slf4j.org/download.html#content) version.
> 
> > SLF4J API version 2.0.0 relies on the [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) mechanism to find its logging backend. See the relevant [FAQ entry](https://www.slf4j.org/faq.html#changesInVersion200) for more details.
> 
> does this impact jenkins and all plugins too?

Jetty switched from it's own rather crusty and bug ridden logging facade to slf4j.
Version 2.0.0 is the first version to support JPMS, which Jetty 10 and Jetty 11 also support.
Our use of slf4j 2.0.0 is light weight enough that you can use slf4j 1.7.x as a replacement (we don't use the fluent logging API in slf4j 2.0.0).
This is what the Eclipse RT (and IDE) folks are doing.
Just know that if you come into it this way (by using slf4j 1.7.x instead of 2.0.0), you have a broken JPMS module system.

joakime: Also note that dropwizard installations using Jetty 10 or Jetty 11 are sticking with slf4j 2.0.0 and have had no complaints.

The label "unstable / experimental" with slf4j 2.0.0 refers to the new APIs (which Jetty doesn't use) and internal impl lookups (only Jetty's optional `jetty-slf4j-impl` uses this) as those are still evolving.
The actual implementations surrounding 2.0.0 are quite stable at runtime, many millions of Jetty 10 and Jetty 11 installations use exactly this setup. (common setups are sfl4j2 to logback, or log4j2, or java.util.logging impls)

olamy: I would not be scare about the version naming of slf4j.
alpha5 has been here since august 2021 and alpha6 has very minor changes.
It's just using a more conservative naming approach/release strategy and not release every changes every day.

olamy: Anyway not sure we really need jpms friendly for Jenkins. So slf4j older versions is an option. And to align with the version range used by Jenkins core. (And being a bit conservative)
I don't really have strong opinions here.

Jenkinsfile
timja: > Java version: 17.0.2, vendor: Eclipse Adoptium

lgtm

pom.xml
```diff
@@ -73,6 +74,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+<!--          <systemPropertyVariables>-->
```
timja, line 76: this needed?

timja (approved):
I merged in master after the conflict in https://github.com/jenkinsci/winstone/pull/203

one q below

pom.xml
olamy: it's just a comment and it has been useful for debugging why jetty didn't start but as I didn't commit the jul.properties.
does it really hurt to a have a comment in a pom.xml?

timja: you did commit it 😄.

Is it useful enough to always have on?

what about
```suggestion
```suggestion
<!-- Uncomment to enable extra logging for debugging -->
<!--          <systemPropertyVariables>-->
```
```

Comment (olamy) [reply pom.xml:77]: done

File: src/main/java/winstone/Launcher.java (hunk 7 above)
Comment (olamy) [reply src/main/java/winstone/Launcher.java]: you asked for a separate PR so automatically it will be an entry in the release notes.

Comment (basil) [reply src/main/java/winstone/Launcher.java]: > automatically it will be an entry in the release notes

Not if one of [four exclusion labels](https://github.com/jenkinsci/.github/blob/4161800c96ca9cd294b5610b96995ee708a6ce9c/.github/release-drafter.yml#L52-L56) are used.

But that does not answer my question. You wrote that this _deserves_ an entry in the release notes (implying the exclusion labels should not be used). Why does this _deserve_ an entry in the release notes?

Comment (olamy) [reply src/main/java/winstone/Launcher.java]: > > automatically it will be an entry in the release notes
> 
> Not if one of [four exclusion labels](https://github.com/jenkinsci/.github/blob/4161800c96ca9cd294b5610b96995ee708a6ce9c/.github/release-drafter.yml#L52-L56) are used.
> 
> But that does not answer my question. You wrote that this _deserves_ an entry in the release notes (implying the exclusion labels should not be used). Why does this _deserve_ an entry in the release notes?

can we move forward please. I definitely have better to do in my life.

File: .github/dependabot.yml [hunk 9]
```diff
@@ -8,7 +8,7 @@ updates:
     ignore:
       # Restrict updates to jetty in the 9.4.x space
```
basil, line 9: Now that we are ignoring versions greater than or equal to 11 rather than greater than or equal to 10, this comment is now incorrect and needs to be updated:

```suggestion
      # Restrict updates to jetty in the 10.x space
```

README.md
```diff
@@ -85,6 +85,8 @@ To run different web applications for diffent virtual hosts:
          / --httpsCertificate     file and the corresponding certificate file
        --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)
        --http2Port              = set the http2 listening port. -1 to disable, Default is disabled
+       --httpsSniHostCheck     = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
```
basil, line 88: The equals sign is not aligned with the other equals signs on the preceding line and the next line:

```suggestion
       --httpsSniHostCheck      = if the SNI Host name must match when there is an SNI certificate. Check disabled per default
```

pom.xml
```diff
@@ -19,7 +19,7 @@
   <properties>
     <revision>6.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jetty.version>9.4.48.v20220622</jetty.version>
+    <jetty.version>10.0.9</jetty.version>
```
basil: This version is a few versions behind. I suggest using the latest version:

```suggestion
    <jetty.version>10.0.11</jetty.version>
```

pom.xml
```diff
@@ -159,6 +163,24 @@
...
+                <enforceBytecodeVersion>
+                  <excludes combine.children="append">
+                    <exclude>org.eclipse.jetty*:*</exclude>
+                  </excludes>
+                </enforceBytecodeVersion>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
basil, line 183: This exclusion is no longer necessary as of parent POM 1.77 (which is in use a few lines above), so I suggest it be deleted:

```suggestion
```

pom.xml
olamy: already done. maybe just wait I mark it ready for review.

basil (commented):
As of this PR, the Winstone JAR now contains new `.class` files in `org/objectweb/asm/` (e.g., `org/objectweb/asm/AnnotationVisitor.class`) and `org/slf4j/` (e.g., `org/slf4j/MDC.class`).

Core already contains ASM and SLF4J JARs in `WEB-INF/lib`. So unfortunately this PR puts us in a position where we now deliver two copies of the ASM and SLF4J classes: one copy in `winstone.jar` in `jenkins.war`, and one copy in `WEB-INF/lib/{asm,slf4j}*.jar` in `jenkins.war`. I think this carries a risk of class loading conflicts due to duplicate classes on the class path. To avoid this risk, I suggest [relocating these classes](https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html) in the shaded artifact in order to create a private copy of their bytecode.

olamy:
uhm regarding asm that's coming from jetty-annotations via websocket.
I'm not sure we need it here. Are websockets from remoting created via annotations?

basil:
> Are websockets from remoting created via annotations?

No., we are not creating WebSockets from Remoting via annotations. I agree that it would be preferable to exclude the ASM library if it is not needed than to relocate the package in the shaded JAR, as this reduces complexity and therefore the surface area of the risk.

olamy:
`jetty-annotations` have been excluded.
Please note now compared there is no more transitive inclusion of websocket-client (which was a mistake in jetty 9.4).
But remoting is using a different one (e.g Tyrus) https://github.com/jenkinsci/remoting/blob/b09e7ead7a03ce69b92a2233e8dc6dc0a0e5abc6/pom.xml#L83

olamy:
fix in jetty sources regarding annotations dependency not needed https://github.com/eclipse/jetty.project/pull/8199

pom.xml
```diff
@@ -313,4 +327,16 @@
...
+    <profile>
+      <id>jdk17</id>
+      <activation>
+        <jdk>17</jdk>
+      </activation>
+      <properties>
+        <argLine>--add-opens=java.base/sun.security.util=ALL-UNNAMED</argLine>
+      </properties>
+    </profile>
+  </profiles>
```
timja, line 340: is this still needed? or was the need for it removed in https://github.com/jenkinsci/winstone/pull/203

timja (commented):
looks good, I think we should run it through the tests in Jenkins core and likely Bom and ATH before shipping? Does that sound reasonable?

olamy:
> looks good, I think we should run it through the tests in Jenkins core and likely Bom and ATH before shipping? Does that sound reasonable?

correct.
I'm still not sure if we really to shade slf4j. I will check.
There is normally enough isolation between jetty core classloader and the webapp classloader (that's something part the ee specs) but it's not really complicated.
anyway if you can test an incremental build version of this PR that would be nice.

basil:
> to avoid scaring too much people :rofl:

I am not scared of class loading.

> I'm still not sure if we really need to shade slf4j. I will check.

I am not 100% sure either, so I am glad you are checking. The problem, in a nutshell, is that Jetty 10 has grown a dependency on SLF4J. We currently deliver SLF4J, with JUL bindings, to plugins via `WEB-INF/lib/slf4j*.jar` in `jenkins.war`, and Jetty 9 does not use SLF4J in our configuration. So we have to ensure that existing plugins continue to work, while also delivering SLF4J to Jetty 10 with JUL bindings. The current PR does the latter by including a copy of the SLF4J classes in a the shaded `winstone.jar` in `jenkins.war`, which duplicates the same classes in `WEB-INF/lib/slf4j*.jar` in `jenkins.war` that we use to provide SLF4J for plugins.

Based on my reading of https://www.eclipse.org/jetty/documentation/jetty-10/programming-guide/index.html#pg-server-http-handler-use-webapp-context it seems that `org.eclipse.jetty.webapp.WebAppClassLoader` is child-first rather than parent-first:

> When `WebAppClassLoader` is asked to load a class, it first tries to find the class locally (since it must use the inverted child-first model); if the class is found, and it is not a system class, the class is loaded; otherwise the class is not found locally. If the class is not found locally, the parent class loader is asked to load the class; the parent class loader uses the standard parent-first model, so it delegates the class loading to its parent, and so on. If the class is found, and it is not a server class, the class is loaded; otherwise the class is not found and a `ClassNotFoundException` is thrown.

So based on my reading of this, a plugin that uses SLF4J would try to load the class from `WebAppClassLoader`, which would find the class in `WEB-INF/lib` and not see the one from Winstone. That is fine. Now, I looked into the definition of "server class" and "system class" in Jetty 10, and I found that Java SE and Jakarta classes are "system classes" while Jetty classes are "server classes." So since the transitive closure of server classes is hidden from `WebAppClassLoader`, and since that transitive closure includes SLF4J, then I think we are safe. Note that the definitions of "system class" and "server class" are different in Jetty 9, though!

While I still think that relocating the Winstone copy of SLF4J could not hurt and would eliminate any possibility of linkage errors, the relocation process carries its own complexity and risk, so I do not feel strongly either way. But I do feel strongly that we should verify my reasoning from above is accurate by empirically stepping through the class loading in a debugger in two scenarios: loading SLF4J from within Jetty, and loading SLF4J from within a Jenkins plugin. Both copies should be loaded separately, in separate class loaders: the system class loader for the former, and `WebAppClassLoader` for the latter. Furthermore, trying to load a Jetty class that itself uses SLF4J from the context of a Jenkins plugin should fail because it is a "server class." If we verify all of this empirically in a debugger, then my concerns about SLF4J would be completely addressed. If these tests were performed, I would be so convinced of the isolation that I would not even be concerned that Jetty is using a newer version of SLF4J than core is providing to plugins.

basil:
> you really should not use bouncycastle inside core or any library it uses. bouncycastle comes from a plugin and having it in core will create issues

James, it is a little more complicated than that with Jetty, which has its own concept of "server classes" and "system classes" with varying degrees of visibility to the underlying web application class loader. Having spent some time researching and writing https://github.com/jenkinsci/winstone/pull/181#issuecomment-1165032477 I do actually think it could be possible to have Bouncy Castle visible to Jetty but not to core and plugins, by configuring it as a "server class" in the Jetty 10 configuration. Whether this is worth somebody's time to implement is a different story.

> mark the feature as deprecated (in the help and release notes?) (but do not remove it) and skip the test on java17 (so long as it compiles?) (ie kick that can down the road for a few years :) )

Actually it has been deprecated and logging a deprecation warning since 2016.

joakime:
The behavior of Server and System classes in Jetty hasn't changed from Jetty 6 thru Jetty 12.
Only the means to configure them has.

In Jetty 9 you configured by package/class.
In Jetty 10 you can configure by package/class/location/module.

Assuming you have the classes on the Server ClassLoader, and are using `.setParentLoaderPriority(false)` (the default value)...

Each entry for Server and System ClasspathPattern can have a positive/inclusive entry (`"entry"`) or negative/exclusive entry (`"-entry"`).

An entry can be ...

* A Package - `"com.foo."` (be sure to end this in a `.`)
* A Location - `"file:///path/to/my/lib/"` (be sure to start this with a `file` scheme and end it in a `/`, must be [an absolute URI reference](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URI.html#isAbsolute()))
* A Module - `"jrt://my.module"` (this must start with `jrt` scheme and be [an absolute URI reference](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URI.html#isAbsolute()))
* A Class - `"com.foo.Bar"` (this must be a FQCN, and is the default assumption for a `ClasspathPattern` entry if it doesn't fit the choices above)

Behaviors for entries ...

* If you specify a positive Server ClasspathPattern entry then it's prevented from being seen by the WebApp
* If you specify a negative Server ClasspathPattern entry then it's available to the WebApp.
* If you specify a positive System ClasspathPattern entry then the WebApp is forced to use that class from the Server ClassLoader.
* If you specify a negative System ClasspathPattern entry then the WebApp cannot see that class at the WebApp. (this mode isn't practical to use)

We have a demo project that can show these behaviors with regards to logging.

https://github.com/jetty-project/logging-overlap-server-webapp

basil:
> The behavior of Server and System classes in Jetty hasn't changed from Jetty 6 thru Jetty 12.
> Only the means to configure them has.

Yeah, I noticed that while comparing https://github.com/eclipse/jetty.project/blob/1f4c5313576a5d02e7906db0d7cf0a0aa2a40ef3/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/WebAppContext.java#L106-L158= with https://github.com/eclipse/jetty.project/blob/2dfbea1bc4e59de650920d4e915edb834f70098d/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/WebAppContext.java#L163-L179=

The Jetty 10 code more clearly indicates to me that anything Jetty related is hidden from the application. For us, that is important, because loading a Jetty class could transitively load an SLF4J class from Jetty's copy of SLF4J, and we don't want that copy of SLF4J to ever be exposed to a Jenkins plugin, which should only ever see SLF4J from the Jenkins core class loader (i.e., from `WEB-INF/lib`). It is not quite as clear to me that the same would apply in Jetty 9, because `__dftServerClasses` includes e.g. `org.eclipse.jetty.server.session.SessionData` but not other classes in `org.eclipse.jetty.server.session`. But that is not a problem in Jetty 9, because no Jetty 9 classes use SLF4J.

I also note that `org.objectweb.asm.` is part of the list in Jetty 9 but not Jetty 10, so it is a good thing we caught the inclusion of ASM and are excluding it in this PR.

basil:
> We have a demo project that can show these behaviors with regards to logging. https://github.com/jetty-project/logging-overlap-server-webapp

Wow! Well, the existence of

> this is the `jetty.base` server libs for slf4j, Using a version newer then what the webapp has.

is more proof than I ever could have needed. So suffice it to say I don't think there is a need for us to verify this behavior in a debugger for Jenkins. It is already covered by this demo project.

basil (approved):
As of my latest two commits, the PR build is passing, and I also verified that the shaded JAR looks good with no unexpected classes. ASM is not present, and SLF4J is present as expected. From my perspective, the next step is to test this in the Jenkins core test suite and PCT/ATH tests as suggested by Tim.

joakime:
@basil don't compare the default lists, that's a deceptive comparison.

In Jetty 9 the lists is hardcoded and occasionally modified.
In Jetty 10 the list is defaulted, and ALWAYS modified.

The modifications come from the various `org.eclipse.jetty.webapp.Configuration` that execute.
Example:

https://github.com/eclipse/jetty.project/blob/jetty-10.0.11/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/WebAppConfiguration.java

``` java
    public WebAppConfiguration()
    {
        addDependencies(WebXmlConfiguration.class, MetaInfConfiguration.class, WebInfConfiguration.class);
        addDependents(JettyWebXmlConfiguration.class);
        protectAndExpose(
            "org.eclipse.jetty.servlet.StatisticsServlet",
            "org.eclipse.jetty.servlet.DefaultServlet",
            "org.eclipse.jetty.servlet.NoJspServlet"
        );
        expose("org.eclipse.jetty.servlet.listener.");
    }
```

If you want a valid comparison, set the `Server.setDumpAfterStart(true)` in both cases (Jetty 9 and Jetty 10).
The output from that will show you the in-use ServerClasses and SystemClasses for each environment.
Use that as a base of your comparison.

basil:
PCT passed: https://ci.jenkins.io/job/Tools/job/bom/job/PR-1226/3/

pom.xml
```diff
@@ -280,12 +284,22 @@
...
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-jmx</artifactId>
-      <version>${jetty.version}</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty.websocket</groupId>
-      <artifactId>websocket-server</artifactId> <!-- or javax-websocket-server-impl -->
-      <version>${jetty.version}</version>
+      <artifactId>websocket-jetty-server</artifactId> <!-- or websocket-javax-server -->
+      <!-- until Jetty release including https://github.com/eclipse/jetty.project/pull/8199 -->
```
jglick, line 291: note: now https://github.com/eclipse/jetty.project/pull/10316 (so far unreleased)

PR #182, opened 2021-12-19
Run build on Java 11

basil:
This can be merged today, because it doesn't bump the `java.level`. I tested this locally, though note that the PR build will not include these changes because I do not have write access to this repository.

CC @olamy

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ properties([[$class: 'BuildDiscarderProperty',
  *  https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
  */
 Map branches = [:]
-['maven'/* TODO broken , 'maven-windows'*/].each {label ->
+['maven-11'/* TODO broken , 'maven-windows'*/].each {label ->
     branches[label] = {
         node(label) {
             stage('Checkout') {
@@ -30,7 +30,7 @@ Map branches = [:]
                     /* Archive the test results */
                     junit '**/target/surefire-reports/TEST-*.xml'
 
-                    if (label == 'maven') {
+                    if (label == 'maven-11') {
                         infra.prepareToPublishIncrementals()
                     }
                 }


Jenkinsfile
```diff
@@ -9,7 +9,7 @@ properties([[$class: 'BuildDiscarderProperty',
  *  https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
  */
 Map branches = [:]
-['maven'/* TODO broken , 'maven-windows'*/].each {label ->
+['maven-11'/* TODO broken , 'maven-windows'*/].each {label ->
```
olamy, line 12: but we will not build master with java 8... until we are jetty 9 and jenkins still java 8 we need to keep the build using java 8...

olamy (changes requested):

Jenkinsfile
basil: Not true at all, Olivier. As long as `java.level` is set to 8, the Java 11 compiler will build the appropriate bytecode for existing use cases. Animal Sniffer enforces that we don't accidentally use newer APIs.

olamy: I know but tests will be run using java 11.  and winstone have some weird code using some reflection on class from package `sun.security` . this is definitely working with java 11 but who knows...
I will definitely be happy to move to java 11 but until core still require java 8 I prefer we run the tests with java 8.
maybe a matrix?

basil: > maybe a matrix?

Running the `Jenkinsfile` build with both Java 8 and Java 11 makes sense to me. That is what most plugins and core do already. But how could I accommodate your request without write access to this repository? Any changes I make to the `Jenkinsfile` in this PR won't get executed because of my lack of write access. So I suggest either(a) you update the `Jenkinsfile` to test against both Java 8 and Java 11 (a PR which you, as the maintainer with write access, could test and iterate on), or (b) you give me write access so I can test the changes you have requested in this PR.

oleg-nenashev: > But how could I accommodate your request without write access to this repository

As a Jenkins core maintainer, you should have access to this repository.
If @olamy does not mind, I will update the permissions accordingly

basil: Thanks @oleg-nenashev, I'm still interested in getting the permissions updated if possible.

Jcmarc1flores (approved):

Jcmarc1flores (approved):

Jenkinsfile
timja: Ping @olamy, / @jglick shall we give core write access?

olamy: That's fine as long as this doesn't turn into crazy version bump which makes this non usable...

timja: I’ve updated the permissions for core

jtnord: > Not true at all, Olivier. As long as `java.level` is set to 8, the Java 11 compiler will build the appropriate bytecode for existing use cases. Animal Sniffer enforces that we don't accidentally use newer APIs.

Code should not rely on animal sniffer for JDK classes, rather use `-release` flag when on JDK9 or higher.  
There are also multi-release jars in any dependencies that will not be flagged using animal sniffer

jtnord (changes requested):
code should not build with java11 unless it is strictly configuring the the maven-compiler plugin to use the `-release` option when run on JDKs higher than 8 (ie 9 and above)

At the same time - I do not see what benefit using java11 here brings?

Jenkinsfile
timja: In the plugin pom this appears to be: https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml#L959-L974

Not present in 'pom' though

jtnord, line 12: BTW - would it not make sense to keep java8 as an option here?  Jenkins still supports java8 for the LTS versions and if there are any security issues in Jetty that need an upgrade it would be nice to know we have tested winstone in a Java8 environement (and for non LTS also until java11 becomes a strict requirement)

```suggestion
['maven-11', 'maven' /* TODO broken , 'maven-windows'*/].each {label ->
```

basil:
> code should not build with java11 unless it is strictly configuring the the maven-compiler plugin to use the `-release` option when run on JDKs higher than 8 (ie 9 and above)

Did you not read my previous reply to Olivier?

Jenkinsfile
basil: > That's fine as long as this doesn't turn into crazy version bump which makes this non usable...

Thank you for the vote of confidence, Olivier.

jtnord:
> > code should not build with java11 unless it is strictly configuring the the maven-compiler plugin to use the `-release` option when run on JDKs higher than 8 (ie 9 and above)
> 
> Did you not read my previous reply to Olivier?

If you mean 

> Not true at all, Olivier. As long as java.level is set to 8, the Java 11 compiler will build the appropriate bytecode for existing use cases. Animal Sniffer enforces that we don't accidentally use newer APIs.

Then yes I did, I am pointing out that is not fully correct and should be adjusted (Tim even linked to the lines of code how to do this)
For the second part Tim updated the permissions so you have write access, so Jenkinsfile changes will be tested..
What did I miss?

basil:
> What did I miss?

The part where I wrote:

> Running the `Jenkinsfile` build with both Java 8 and Java 11 makes sense to me.

In other words, I agreed with Olivier (and, by extension, your future comments). That is why, once I attained write access, I went ahead and implemented Olivier's original suggestion and closed this PR.

Jenkinsfile
olamy: > > That's fine as long as this doesn't turn into crazy version bump which makes this non usable...
> 
> Thank you for the vote of confidence, Olivier.

if you could give some time for people to review PRs that would be great.

olamy: @timja @basil it's supposed to be a project with people working together? so it turns now to something to a one man show doing changes and not even waiting for approval on pull requets.
what a shame!!!

basil:
> If you could give some time for people to review PRs that would be great.

@olamy Thank you for expressing your interest in reviewing PRs! I was not aware that you were interested in doing reviews, and by the time I received your messages the PRs had already been merged. I am sorry if you were upset by this. I will give future PRs a waiting period in order to allow for you (or anyone else) to review them.

> so it turns now into something to a one man show doing changes and not even waiting for approval

I waited for approval in #193, #195, #197, #198, #200, all of which contained non-trivial changes. I did not wait for approval for #194, #196, #199, #201, and #202, all of which contained trivial changes. If I acted in error, feel free to revert and I can open a new PR for further discussion.

> what a shame!!!

May I remind you about the [Jenkins Code of Conduct](https://www.jenkins.io/project/conduct/). Please act with kindness and refrain from public shaming. As you wrote, this is "a project with people working together." Assuming good faith goes a long way.

olamy:
> 
> May I remind you about the [Jenkins Code of Conduct](https://www.jenkins.io/project/conduct/). Please act with kindness and refrain from public shaming. As you wrote, this is "a project with people working together." Assuming good faith goes a long way.

please don't go this way but looks at what happened.....
code of conduct means asking review from long time committers on a project.....
and not changing everything suddenly... because you prefer a different style of code....
and it's what happened here... 
you cannot wait only few minutes for PR validation... (this is what I mean with working together)
it's part of an implicit code of conduct...

Jenkinsfile
timja: Not sure why I’m tagged here…

PR #200, opened 2022-02-07
Fix high-threshold SpotBugs warnings

basil:


src/main/java/winstone/AbstractSecuredConnectorFactory.java
```diff
@@ -122,10 +122,10 @@ protected void configureSsl( Map<String, String> args, Server server ) throws IO
     }
 
 
-    private static PrivateKey readPEMRSAPrivateKey(Reader reader) throws IOException, GeneralSecurityException {
+    private static PrivateKey readPEMRSAPrivateKey(File opensslKey) throws IOException, GeneralSecurityException {
         // TODO: should have more robust format error handling
         try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
-            try (BufferedReader r = new BufferedReader( reader )) {
+            try (BufferedReader r = Files.newBufferedReader(opensslKey.toPath(), StandardCharsets.US_ASCII)) {
```
basil, line 128: This is Base64-encoded per [RFC3447 - Appendix A.1.2](https://www.rfc-editor.org/rfc/rfc3447#appendix-A.1.2). See also the Base64 decode call a few lines below. Since this should not contain any special characters, ASCII should suffice.

src/main/java/winstone/Launcher.java
```diff
@@ -442,15 +449,23 @@ public static void initLogger(Map<String, String> args) throws IOException {
         // boolean showThrowingLineNo = Option.LOG_THROWING_LINE_NO.get(args);
         boolean showThrowingThread = Option.LOG_THROWING_THREAD.get(args);
         OutputStream logStream;
+        Charset logCharset;
         if (args.get("logfile") != null) {
-            logStream = new FileOutputStream(args.get("logfile"));
+            try {
+                logStream = Files.newOutputStream(Paths.get(args.get("logfile")), StandardOpenOption.CREATE, StandardOpenOption.APPEND);
```
basil, line 455: Technically appending to an existing log file is a change in behavior, but I think it makes more sense to append. In `jenkinsci/packaging` it is currently recommended to set the `-logfile` option to `/var/log/jenkins/jenkins.log`, where it will be rotated by `logrotate(8)`.

src/main/java/winstone/Launcher.java
```diff
@@ -442,15 +449,23 @@ public static void initLogger(Map<String, String> args) throws IOException {
...
+            } catch (InvalidPathException e) {
+                throw new IOException(e);
+            }
+            logCharset = StandardCharsets.UTF_8;
```
basil, line 459: Technically a change in behavior, but I think it makes the most sense for this log to be in UTF-8.

pom.xml
```diff
@@ -24,7 +24,7 @@
     <alpn.api.version>1.1.3.v20160715</alpn.api.version>
     <java.level>8</java.level>
     <!-- TODO: remove once SpotBugs issues are fixed. 57 so far -->
```
jglick, line 26: Still valid?

src/main/java/winstone/accesslog/SimpleAccessLogger.java
```diff
@@ -78,8 +81,12 @@ public SimpleAccessLogger(String webAppName, Map<String, String> startupArgs)
...
             Logger.logDirectMessage(Logger.WARNING, null, "Failed to mkdirs " + parentFile.getAbsolutePath(), ex);
         }
-        this.outStream = new FileOutputStream(file, true);
-        this.outWriter = new PrintWriter(this.outStream, true);
+        try {
+            this.outStream = Files.newOutputStream(file.toPath(), StandardOpenOption.CREATE, StandardOpenOption.APPEND);
+        } catch (InvalidPathException e) {
+            throw new IOException(e);
+        }
+        this.outWriter = new PrintWriter(new OutputStreamWriter(this.outStream, StandardCharsets.UTF_8), true);
```
basil, line 89: Technically a change in behavior, but I think it makes the most sense for this log to be in UTF-8.

jglick (approved):

pom.xml
basil: 29 issues left now :-) I'll update the number.

olamy:
could we have a bit of time to review???

src/main/java/winstone/accesslog/SimpleAccessLogger.java
olamy: as it's a breaking change the PR should have the appropriate label...

src/main/java/winstone/Launcher.java
olamy: as it's a breaking change the PR should have the appropriate label...

src/main/java/winstone/Launcher.java
olamy: as it's a breaking change the PR should have the appropriate label...

basil: > as it's a breaking change the PR should have the appropriate label...

I updated the label and the changelog to mention this change.

src/main/java/winstone/Launcher.java
basil: > as it's a breaking change the PR should have the appropriate label...

I updated the label and the changelog to mention this change.

src/main/java/winstone/accesslog/SimpleAccessLogger.java
basil: > as it's a breaking change the PR should have the appropriate label...

I updated the label and the changelog to mention this change.

PR #203, opened 2022-02-08
Remove automatic self signed cert if tls specified but no keystore

timja:
see https://github.com/jenkinsci/winstone/pull/181#issuecomment-1032544226

### DIFF
--- a/src/main/java/winstone/AbstractSecuredConnectorFactory.java
+++ b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
@@ -29,7 +29,6 @@ import java.security.KeyStore;
 import java.security.PrivateKey;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateFactory;
-import java.security.cert.X509Certificate;
 import java.security.spec.RSAPrivateKeySpec;
 import java.text.MessageFormat;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


basil (approved):
These unused imports can now be removed:

```java
diff --git a/src/main/java/winstone/AbstractSecuredConnectorFactory.java b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
index 9a54079..7b8d55a 100644
--- a/src/main/java/winstone/AbstractSecuredConnectorFactory.java
+++ b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
@@ -29,7 +29,6 @@ import java.security.KeyStore;
 import java.security.PrivateKey;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateFactory;
-import java.security.cert.X509Certificate;
 import java.security.spec.RSAPrivateKeySpec;
 import java.text.MessageFormat;
 import java.util.Arrays;
diff --git a/src/test/java/winstone/HttpsConnectorFactoryTest.java b/src/test/java/winstone/HttpsConnectorFactoryTest.java
index 9b43490..72ab567 100644
--- a/src/test/java/winstone/HttpsConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpsConnectorFactoryTest.java
@@ -4,7 +4,6 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 
 import org.apache.commons.io.IOUtils;
 import org.eclipse.jetty.server.HttpConnectionFactory;
@@ -14,7 +13,6 @@ import org.junit.Test;
 
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.X509TrustManager;
 import java.net.HttpURLConnection;
 import java.net.URL;
```

timja:
Spotless? :)

basil:
> Spotless? :)

Sure, but seems laborious to set it up from scratch in each individual repository. I've been meaning to look into extracting Spotless/Checkstyle configuration into `jenkinsci/{plugin-}pom` so that individual components or plugins can opt into it with a common set of rules, but that is a bigger project and certainly off-topic for this PR.

jglick (approved):
OK I guess. No particular opinion since I normally deal with K8s installations in which TLS is terminated at the ingress.

olamy (approved):
definitely good to remove this non safe "feature".
but who knows who is using that as it was an automatic "feature" so some people may use this without knowing it. I suspect this will create some issues but there is an exception so we have to hope people test their upgrade before doing it 😂.. (2ezy now to click on dependabot PR merge button without more testing.

jglick:
I think it was a useful system for testing, though something like https://github.com/jglick/jenkins-demo-reverse-proxy works without it. No idea how many people were really using it in production.

timja:
I'll ship just this change, so we can see if we get some feedback on it. I think unlikely though...

PR #205, opened 2022-02-08
Add additional note on loopback address to test

timja:
It's in the README but I missed it, if it had been in the test I think I would have found the issue without having to google for a bit

### DIFF
--- a/src/test/java/winstone/HttpConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpConnectorFactoryTest.java
@@ -22,6 +22,7 @@ public class HttpConnectorFactoryTest extends AbstractWinstoneTest {
         args.put("warfile", "target/test-classes/test.war");
         args.put("prefix", "/");
         args.put("httpPort", "0");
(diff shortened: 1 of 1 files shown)


basil (approved):

MarkEWaite (approved):

src/test/java/winstone/HttpConnectorFactoryTest.java
```diff
@@ -22,6 +22,7 @@ public void testListenAddress() throws Exception {
         args.put("warfile", "target/test-classes/test.war");
         args.put("prefix", "/");
         args.put("httpPort", "0");
+        // see https://superuser.com/a/458877/714797 for getting this to work on macOS
```
olamy, line 25: `just see README development part` as it's already there.
no need to link to an external website which has exactly the same content as a file already here

timja:
```suggestion
        // see README development section for getting this to work on macOS
```

olamy (approved):

PR #206, opened 2022-02-08
Remove `commonLibCLPaths`

basil:
I could find no references to `commonLibCLPaths` [anywhere in the Jenkins ecosystem](https://github.com/search?type=Code&q=user%3Ajenkinsci+commonLibCLPaths`), let alone in this repository. If you look at the deleted code, we are just populating this array and never consuming it, so it is dead code. I could find no usages of the `HostConfiguration` or `HostGroup` constructors outside of this repository, either. This dead code seems best left in the rear-view mirror.

### DIFF
--- a/src/main/java/winstone/HostConfiguration.java
+++ b/src/main/java/winstone/HostConfiguration.java
@@ -49,18 +49,16 @@ public class HostConfiguration {
     private Map<String, String> args;
     private Map<String,WebAppContext> webapps;
     private ClassLoader commonLibCL;
-    private File[] commonLibCLPaths;
     private MimeTypes mimeTypes = new MimeTypes();
     private final LoginService loginService;
 
     public HostConfiguration(Server server, String hostname, ClassLoader commonLibCL,
-                             File[] commonLibCLPaths, Map<String, String> args, File webappsDir) throws IOException {
+                             Map<String, String> args, File webappsDir) throws IOException {
         this.server = server;
         this.hostname = hostname;
         this.args = args;
         this.webapps = new Hashtable<>();
         this.commonLibCL = commonLibCL;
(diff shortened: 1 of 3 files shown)


timja (approved):

olamy:
why not adding new constructors without this and leave those constructors `@deprecated` for 1 or 2 releases.
this is usually what a good practice.

basil:
> why not adding new constructors without this and leave those constructors `@deprecated` for 1 or 2 releases

Because, as I wrote in the review description:

> I could find no usages of the `HostConfiguration` or `HostGroup` constructors outside of this repository.

olamy:
> > why not adding new constructors without this and leave those constructors `@deprecated` for 1 or 2 releases
> 
> Because, as I wrote in the review description:
> 
> > I could find no usages of the `HostConfiguration` or `HostGroup` constructors outside of this repository.

I agree the probability is very low but who knows, the world is not only what is in github repositories.

basil:
If you are able to show that there is a usage of the old constructor, I am willing to mark it as `@Deprecated`. If you are not able to show that there is a usage of the old constructor, I am not willing to mark it as `@Deprecated`.

olamy:
> If you are able to show that there is a usage of the old constructor, I am willing to mark it as `@Deprecated`. If you are not able to show that there is a usage of the old constructor, I am not willing to mark it as `@Deprecated`.

this is not exactly in this way we should proceed such change.
but LOL you really made my day here...

olamy:
@basil why closing that? I'm not against removing/cleaning those unused references that's perfectly fine. 
I'm just asking to be cautious.

basil:
> @basil why closing that?

@olamy As I wrote previously:

> If you are not able to show that there is a usage of the old constructor, I am not willing to mark it as `@Deprecated`.

PR #229, opened 2022-06-22
Make test less flaky

basil:
Before adding this sleep, this test would flake about half the time as in [this run](https://ci.jenkins.io/job/Core/job/winstone/job/master/252/testReport/winstone.accesslog/SimpleAccessLoggerTest/testSimpleConnection/). After adding this sleep I have run the test 10 times in a row without any flakes.

### DIFF
--- a/src/test/java/winstone/accesslog/SimpleAccessLoggerTest.java
+++ b/src/test/java/winstone/accesslog/SimpleAccessLoggerTest.java
@@ -40,6 +40,7 @@ public class SimpleAccessLoggerTest extends AbstractWinstoneTest {
         makeRequest("http://localhost:"+port+"/examples/CountRequestsServlet");
 
         // check the log file
+        Thread.sleep(5000);
         String text = FileUtils.readFileToString(logFile, StandardCharsets.UTF_8);
         assertEquals(String.format("127.0.0.1 - - GET /examples/CountRequestsServlet HTTP/1.1 200%n"),text);
     }


src/test/java/winstone/accesslog/SimpleAccessLoggerTest.java
```diff
@@ -40,6 +40,7 @@ public void testSimpleConnection() throws Exception {
         makeRequest("http://localhost:"+port+"/examples/CountRequestsServlet");
 
         // check the log file
+        Thread.sleep(5000);
```
olamy, line 43: would be better to add a loop retrying every maybe 100ms then if keep failing after 5000ms
because this add unconditionally 5s to the build

olamy:
why making a PR if you do not wait for review?

src/test/java/winstone/accesslog/SimpleAccessLoggerTest.java
basil: I am not interested in doing that.

basil:
I am not interested.

olamy:
> I am not interested.

sounds like a polite, technical and community oriented answer....

src/test/java/winstone/accesslog/SimpleAccessLoggerTest.java
timja: recorded in https://github.com/jenkinsci/winstone/issues/231 for future improvement

Issue #213, opened 2022-03-12
Cannot read OpenSSL-style PEM-encoded RSA private key on Java 17


basil:
### Jenkins and plugins versions report

Apache Maven 3.8.4 (9b656c72d54e5bacbed989b64718c159fe39b537)
Java version: 17.0.2, vendor: Private Build, runtime: /usr/lib/jvm/java-17-openjdk-amd64
OS name: "linux", version: "5.4.0-104-generic", arch: "amd64", family: "unix"

### What Operating System are you using (both controller, and any agents involved in the problem)?

Ubuntu 20.04.4 LTS x86_64

### Reproduction steps

Run `mvn clean verify -Dtest=winstone.HttpsConnectorFactoryTest` with Java 17.

### Expected Results

**Note:** These are the _actual_ results when running with Java 8 or Java 11.

The test passes.

### Actual Results

The test fails with

```
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.77 s <<< FAILURE! - in winstone.HttpsConnectorFactoryTest
[ERROR] testHttps(winstone.HttpsConnectorFactoryTest)  Time elapsed: 0.255 s  <<< ERROR!
java.io.IOException: Failed to start a listener: winstone.HttpsConnectorFactory
        at winstone.Launcher.spawnListener(Launcher.java:261)
        at winstone.Launcher.<init>(Launcher.java:190)
        at winstone.HttpsConnectorFactoryTest.testHttps(HttpsConnectorFactoryTest.java:42)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #232, opened 2022-06-23
Remove ability to load pem cert for winstone

timja:
Fixes https://github.com/jenkinsci/winstone/issues/213

Test needs re-implementing to use a keystore (that mode wasn't tested before afaict)

## Upgrade guidelines

Support has been removed for PEM encoded certificates when running Jenkins with the embedded Jetty (winstone) container and TLS.
Specifically the flags `--httpsPrivateKey` and `--httpsCertificate` have been removed.

Users who were previously using this feature must use a KeyStore instead.  See the [documentation](https://www.jenkins.io/doc/book/installing/initial-settings/#https-with-an-existing-certificate) for details.

This feature has been logging deprecation warnings on startup since 2016.

### DIFF
--- a/src/main/resources/winstone/LocalStrings.properties
+++ b/src/main/resources/winstone/LocalStrings.properties
@@ -94,8 +94,6 @@ Launcher.UsageInstructions.Options=\
 \   --httpsKeyStore          = the location of the SSL KeyStore file. Default is ./winstone.ks\n\
 \   --httpsKeyStorePassword  = the password for the SSL KeyStore file. Default is null\n\
 \   --httpsKeyManagerType    = the SSL KeyManagerFactory type (eg SunX509, IbmX509). Default is SunX509\n\
-\   --httpsPrivateKey        = this switch with --httpsCertificate can be used to run HTTPS with OpenSSL secret key\n\
-\     / --httpsCertificate     file and the corresponding certificate file\n\
 \   --httpsRedirectHttp      = redirect http requests to https (requires both --httpPort and --httpsPort)\n\
 \   --http2Port              = set the http2 listening port. -1 to disable, Default is disabled\n\
 \   --http2ListenAddress     = set the http2 listening address. Default is all interfaces\n\
@@ -168,9 +166,6 @@ HttpsConnectorFactory.SelfSigned=\
   Creating a self-signed certificate currently relies on unsupported APIs in the Oracle JRE.\n\
   Please create your own certificate using supported tools instead and use --httpsKeyStore.
 HttpsConnectorFactory.SelfSignedError=Failed to create a self-signed certificate; make one yourself.
-HttpsConnectorFactory.LoadPrivateKey=\
-  Using the --httpsPrivateKey/--httpsCertificate options currently relies on unsupported APIs in the Oracle JRE.\n\
-  Please use --httpsKeyStore and related options instead.
 HttpsConnectorFactory.LoadPrivateKeyError=Cannot load private key; try using a Java keystore instead.
 
 Http2ConnectorFactory.FailedStart.ALPN=Failed to start ALPN
--- a/src/main/resources/winstone/LocalStrings.properties
+++ b/src/main/resources/winstone/LocalStrings.properties
@@ -166,7 +166,6 @@ HttpsConnectorFactory.SelfSigned=\
   Creating a self-signed certificate currently relies on unsupported APIs in the Oracle JRE.\n\
   Please create your own certificate using supported tools instead and use --httpsKeyStore.
 HttpsConnectorFactory.SelfSignedError=Failed to create a self-signed certificate; make one yourself.
-HttpsConnectorFactory.LoadPrivateKeyError=Cannot load private key; try using a Java keystore instead.
 
 Http2ConnectorFactory.FailedStart.ALPN=Failed to start ALPN
 
--- a/src/test/java/winstone/HttpsConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpsConnectorFactoryTest.java
@@ -53,7 +53,12 @@ public class HttpsConnectorFactoryTest extends AbstractWinstoneTest {
         args.put("httpsKeyStorePassword", "changeit");
         winstone = new Launcher(args);
         int port = (( ServerConnector)winstone.server.getConnectors()[0]).getLocalPort();
+        assertConnectionRefused("127.0.0.2", port);
         request(new TrustEveryoneManager(), port);
+        LowResourceMonitor lowResourceMonitor = winstone.server.getBean( LowResourceMonitor.class);
+        assertNotNull(lowResourceMonitor);
+        assertFalse(lowResourceMonitor.isLowOnResources());
+        assertTrue(lowResourceMonitor.isAcceptingInLowResources());
     }
 
     @Test
--- a/src/test/java/winstone/HttpsConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpsConnectorFactoryTest.java
@@ -9,7 +9,6 @@ import org.apache.commons.io.IOUtils;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.LowResourceMonitor;
 import org.eclipse.jetty.server.ServerConnector;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import javax.net.ssl.HttpsURLConnection;
--- a/src/test/java/winstone/HttpsConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpsConnectorFactoryTest.java
@@ -30,29 +30,6 @@ import org.jvnet.hudson.test.Issue;
  * @author Kohsuke Kawaguchi
  */
 public class HttpsConnectorFactoryTest extends AbstractWinstoneTest {
-    @Test
-    @Ignore("TODO re-implement with a key store")
-    public void testHttps() throws Exception {
-        Map<String,String> args = new HashMap<>();
-        args.put("warfile", "target/test-classes/test.war");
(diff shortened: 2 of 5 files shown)


src/test/java/winstone/HttpsConnectorFactoryTest.java
```diff
@@ -29,28 +30,6 @@
  * @author Kohsuke Kawaguchi
  */
 public class HttpsConnectorFactoryTest extends AbstractWinstoneTest {
-    @Test
-    public void testHttps() throws Exception {
```
timja, line 33: https is covered by `#wildcard()`

basil: Sure but `wildcard` lacks some of the assertions from `testHttps`, presumably because `testHttps` is (was, as of this PR the "primary" test for TLS and it would be pointless to test the same thing twice. But if the "primary" test for TLS is now going to be `wildcard`, which seems fine, then I think we should move those assertions to `wildcard` to retain existing coverage levels. Otherwise, we would end up with a (slight) regression in test coverage:

```java
diff --git a/src/test/java/winstone/HttpsConnectorFactoryTest.java b/src/test/java/winstone/HttpsConnectorFactoryTest.java
index 72ab567..3a613dc 100644
--- a/src/test/java/winstone/HttpsConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpsConnectorFactoryTest.java
@@ -74,7 +74,15 @@ public class HttpsConnectorFactoryTest extends AbstractWinstoneTest {
         args.put("httpsKeyStorePassword", "changeit");
         winstone = new Launcher(args);
         int port = (( ServerConnector)winstone.server.getConnectors()[0]).getLocalPort();
+        assertConnectionRefused("127.0.0.2", port);
         request(new TrustEveryoneManager(), port);
+        LowResourceMonitor lowResourceMonitor = winstone.server.getBean( LowResourceMonitor.class);
+        assertNotNull(lowResourceMonitor);
+        assertFalse(lowResourceMonitor.isLowOnResources());
+        assertTrue(lowResourceMonitor.isAcceptingInLowResources());
     }
 
     @Test
```

basil (commented):
Some more references in this repository:

```
README.md:         / --httpsCertificate     file and the corresponding certificate file
README.md:       --httpsPrivateKey        = this switch with --httpsCertificate can be used to run HTTPS with OpenSSL secret key
src/main/resources/winstone/LocalStrings.properties:\     / --httpsCertificate     file and the corresponding certificate file\n\
src/main/resources/winstone/LocalStrings.properties:\   --httpsPrivateKey        = this switch with --httpsCertificate can be used to run HTTPS with OpenSSL secret key\n\
src/main/resources/winstone/LocalStrings.properties:  Using the --httpsPrivateKey/--httpsCertificate options currently relies on unsupported APIs in the Oracle JRE.\n\
```

Unused imports:

```java
diff --git a/src/main/java/winstone/AbstractSecuredConnectorFactory.java b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
index d6e3f57..789c342 100644
--- a/src/main/java/winstone/AbstractSecuredConnectorFactory.java
+++ b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
@@ -12,30 +12,16 @@ import org.eclipse.jetty.util.ssl.SslContextFactory;
 import winstone.cmdline.Option;
 
 import javax.net.ssl.KeyManagerFactory;
-import java.io.BufferedReader;
-import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.Method;
-import java.math.BigInteger;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.InvalidPathException;
 import java.security.GeneralSecurityException;
-import java.security.KeyFactory;
 import java.security.KeyStore;
-import java.security.PrivateKey;
-import java.security.cert.Certificate;
-import java.security.cert.CertificateFactory;
-import java.security.spec.RSAPrivateKeySpec;
 import java.text.MessageFormat;
 import java.util.Arrays;
-import java.util.Base64;
 import java.util.Enumeration;
 import java.util.Map;
-import java.util.logging.Level;
```

Some references in documentation that also would need to be removed:

- https://github.com/jenkinsci/docker/blob/1303da3ef5d2057dc28239331ea0ad3dcde1558e/README.md?plain=1#L131=
- https://github.com/jenkins-infra/jenkins.io/blob/bc7c1795fd403ea1f237d62d3ccedbf74f0dc3cc/content/doc/book/installing/_jenkins-command-parameters.adoc?plain=1#L127-L128=

Not hitting the "approve" button just yet because I want to review the removal diff one last time after the imports and strings are taken care of (and I hope to see other PRs in the Docker and jenkins.io repositories), but I am basically ready to approve this change.

I obviously agree that the current implementation cannot be sustained. Removing or reimplementing this is the only obvious way in the long term. And reimplementing something like this ourselves is non-trivial, because we cannot use Bouncy Castle in core. Nor do I feel that reimplementing this is a great use of our time; better to leave crypto for the experts and focus on our strengths. Plus it has been logging a warning since commit 7dd5361 from 2016. Finally, you have already contributed the PR, which is highly appreciated. So I am definitely in favor of removing this and directing users to use the JKS keystore system or another solution, like nginx/Apache.

I do think some users will be caught off guard from this. @markewaite's standard "ask" for such removals is a blog post and an upgrade guide entry. I tend to feel that for minor breaking changes a blog post is usually overkill, but explaining the history, motivation, and migration path in a dedicated piece could not hurt. However I do think an entry in the upgrade guide is necessary in order to avoid bugs from being filed by users who upgrade blindly and lose functionality. I think the minimum information that should be included in the upgrade guide is a recommendation of what to migrate to if you were using this deprecated functionality: in our case, a Java key store (JKS) setup or nginx/Apache.

I realize that asking for docs is not the most pleasant thing in the world, especially after you have already done the thankless work of the removal in the first place, so I am fully prepared to hear "no" for an answer. I myself just finished writing an 8-page blog post about Java 11, so I appreciate the onerous nature of this type of work. If you aren't willing to write an upgrade guide entry yourself, perhaps you could summarize the main points to @kmartens27 and he could give it a shot. If not, Mark and I can probably help as a fallback, but I would rather not (if possible) since I do not really know much about crypto and have to Google it every time (and then promptly forget). Regardless of who does the work, I just want to make sure it doesn't slip through the cracks and that users aren't surprised. So as long as we can come up with a plan for the docs, I am all for this landing this PR.

timja:
I've added an upgrade guide note happy to either take suggestions or for anyone to copy-edit themselves.

src/main/resources/winstone/LocalStrings.properties
```diff
@@ -168,9 +166,6 @@ HttpsConnectorFactory.SelfSigned=\
   Creating a self-signed certificate currently relies on unsupported APIs in the Oracle JRE.\n\
   Please create your own certificate using supported tools instead and use --httpsKeyStore.
 HttpsConnectorFactory.SelfSignedError=Failed to create a self-signed certificate; make one yourself.
-HttpsConnectorFactory.LoadPrivateKey=\
-  Using the --httpsPrivateKey/--httpsCertificate options currently relies on unsupported APIs in the Oracle JRE.\n\
-  Please use --httpsKeyStore and related options instead.
 HttpsConnectorFactory.LoadPrivateKeyError=Cannot load private key; try using a Java keystore instead.
```
basil, line 169: This string is now unused and should also be deleted:

```suggestion
```

basil (approved):
Many thanks for filing the other PRs and writing the upgrade guide entry. Hopefully it goes without saying that I have a deep respect for your thoroughness and your contributions. I copyedited the upgrade guide entry a little to also include the reason behind the removal, which should hopefully soften the blow a little for users who are upset that we are making a breaking change:

> Support for OpenSSL-style PEM-encoded RSA private keys has been removed when running Jenkins with the embedded Jetty (Winstone) container and TLS. Specifically, the `--httpsPrivateKey` and `--httpsCertificate` flags have been removed without replacement. These flags have have printed deprecation warnings since 2016 and were implemented with non-standard APIs that have been removed from Java 17. The recommendation is to migrate to the `--httpsKeyStore` option, which takes a Java KeyStore (JKS) as described in [the documentation](https://www.jenkins.io/doc/book/installing/initial-settings/#https-with-an-existing-certificate).

I have not discussed this removal with Mark, so I do not know if he would have any cause for concern, but I believe the justification is solid: we have to either remove or reimplement this functionality for Java 17, reimplementing it is non-trivial because the obvious implementation with Bouncy Castle is not available to us in core components, and removal is justified by the deprecation notice that has been printed since 2016. From my perspective, this is ready to land. I did note that there are more references to this functionality on the CloudBees support site (https://www.google.com/search?q=%22httpsPrivateKey%22+site%3Asupport.cloudbees.com for some examples), so with Mark's blessing I will file an internal ticket for these references to be updated with the landing of this PR.

MarkEWaite:
> I have not discussed this removal with Mark, so I do not know if he would have any cause for concern, but I believe the justification is solid

I don't have any concern with the removal.  I think that the entry in the upgrade guide is sufficient.  When the core pull request is submitted to merge the winstone release, it would be a great kindness in the proposed upgrade guide text were copied into the Jenkins core pull request.

timja:
> I doubt it would be worthwhile. I do not have an interest in trying to prototype an implementation myself

Agreed, and I do not have the interest at this time, but thanks for the research.

jglick:
Related somewhat to #203?

jglick (approved):
We should not be trying to implement a fully featured web server here. At worst you can run a reverse proxy to terminate TLS. ([demo](https://github.com/jglick/jenkins-demo-reverse-proxy))

PR #249, opened 2022-08-05
Write port value atomically

### LINKED ISSUE #248: Launcher should atomically write port to file
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
 jenkins-test-harness-1816.v8138d8056949
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

MacOs

### Reproduction steps

To reproduce you could run hundreds of time [RemoteArtifactCopyBuilderTest](https://github.com/cloudbees/operations-center-context/blob/master/src/test/java/com/cloudbees/opscenter/artifacts/RemoteArtifactCopyBuilderTest.java#L98) and see them flaky. But keep in mind that you should use jenkins-test-harness <= [1816.v8138d8056949](https://github.com/jenkinsci/jenkins-test-harness/releases/tag/1816.v8138d8056949), as jenkins-test-harness going to patch  [workaround](https://github.com/jenkinsci/jenkins-test-harness/pull/465).

### Expected Results

Tests should not flake

### Actual Results

tests are flaky

### Anything else?

https://github.com/jenkinsci/winstone/blob/4754da67bf797b906a5c858b2de63b97ac609543/src/main/java/winstone/Launcher.java#L221-L242
should write port value atomically.

For a context related conversation https://github.com/jenkinsci/jenkins-test-harness/pull/465#discussion_r936638840

Pldi23:
Fixes #248

### DIFF
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -234,12 +234,13 @@ public class Launcher implements Runnable {
                     Path tmpPath = Files.createTempFile(portDir, "temp", "");
                     try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
-                        try {
-                            Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
-                        } catch (AtomicMoveNotSupportedException e) {
-                            Logger.logDirectMessage(Level.WARNING, "", "Atomic move does not supported", e);
-                            Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
-                        }
+                    }
+
+                    try {
+                        Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
+                    } catch (AtomicMoveNotSupportedException e) {
+                        Logger.logDirectMessage(Level.WARNING, "", "Atomic move does not supported", e);
+                        Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
                     } finally {
                         tmpPath.toFile().deleteOnExit();
                     }
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -239,7 +239,7 @@ public class Launcher implements Runnable {
                     try {
                         Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
                     } catch (AtomicMoveNotSupportedException e) {
-                        Logger.logDirectMessage(Level.WARNING, "", "Atomic move does not supported", e);
+                        Logger.logDirectMessage(Logger.WARNING, null, "Atomic move not supported. Falling back to non-atomic move.", e);
                         Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
                     } finally {
                         tmpPath.toFile().deleteOnExit();
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -246,8 +246,6 @@ public class Launcher implements Runnable {
                             e2.addSuppressed(e);
                             throw e2;
                         }
-                    } finally {
-                        tmpPath.toFile().deleteOnExit();
                     }
                 } else {
                     throw new IllegalStateException("Only ServerConnector is supported");
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -232,9 +232,7 @@ public class Launcher implements Runnable {
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
                     Path tmpPath = Files.createTempFile(portDir, portFile.getFileName().toString(), null);
-                    try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
-                        writer.write(Integer.toString(port));
-                    }
+                    Files.writeString(tmpPath, Integer.toString(port));
 
                     try {
                         Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -10,6 +10,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.jenkins.lib.support_log_formatter.SupportLogFormatter;
 import java.io.BufferedWriter;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Path;
 import org.eclipse.jetty.jmx.MBeanContainer;
 import org.eclipse.jetty.server.Connector;
@@ -233,7 +234,12 @@ public class Launcher implements Runnable {
                     Path tmpPath = Files.createTempFile(portDir, "temp", "");
                     try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
-                        Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
+                        try {
+                            Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
+                        } catch (AtomicMoveNotSupportedException e) {
+                            Logger.logDirectMessage(Level.WARNING, "", "Atomic move does not supported", e);
+                            Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
+                        }
                     } finally {
                         tmpPath.toFile().deleteOnExit();
                     }
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -231,7 +231,7 @@ public class Launcher implements Runnable {
                     Path portFile = Paths.get(portFileName);
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
-                    Path tmpPath = Files.createTempFile(portDir, "temp", "");
+                    Path tmpPath = Files.createTempFile(portDir, portFile.getFileName().toString(), null);
                     try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
                     }
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -36,6 +36,7 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.List;
@@ -229,8 +230,12 @@ public class Launcher implements Runnable {
                     Path portFile = Paths.get(portFileName);
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
-                    try (BufferedWriter writer = Files.newBufferedWriter(portFile, StandardCharsets.UTF_8)) {
+                    Path tmpPath = Files.createTempFile(portDir, "temp", "");
+                    try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
+                        Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
+                    } finally {
+                        tmpPath.toFile().deleteOnExit();
                     }
                 } else {
                     throw new IllegalStateException("Only ServerConnector is supported");
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -240,7 +240,12 @@ public class Launcher implements Runnable {
                         Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
                     } catch (AtomicMoveNotSupportedException e) {
                         Logger.logDirectMessage(Logger.WARNING, null, "Atomic move not supported. Falling back to non-atomic move.", e);
-                        Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
+                        try {
+                            Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
+                        } catch (IOException e2) {
+                            e2.addSuppressed(e);
+                            throw e2;
+                        }
                     } finally {
                         tmpPath.toFile().deleteOnExit();
                     }
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -9,7 +9,6 @@ package winstone;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.jenkins.lib.support_log_formatter.SupportLogFormatter;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Path;
 import org.eclipse.jetty.jmx.MBeanContainer;
 import org.eclipse.jetty.server.Connector;
@@ -33,6 +32,7 @@ import java.net.Socket;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.charset.Charset;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
@@ -231,8 +231,7 @@ public class Launcher implements Runnable {
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
                     Path tmpPath = Files.createTempFile(portDir, portFile.getFileName().toString(), null);
-                    Files.writeString(tmpPath, Integer.toString(port));
-
+                    Files.writeString(tmpPath, Integer.toString(port), StandardCharsets.UTF_8);
                     try {
                         Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
                     } catch (AtomicMoveNotSupportedException e) {
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -8,7 +8,6 @@ package winstone;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.jenkins.lib.support_log_formatter.SupportLogFormatter;
-import java.io.BufferedWriter;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Path;
--- a/src/test/java/winstone/HttpConnectorFactoryTest.java
+++ b/src/test/java/winstone/HttpConnectorFactoryTest.java
@@ -1,7 +1,6 @@
 package winstone;
 
 import java.io.BufferedReader;
-import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -12,8 +11,10 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.FileUtils;
+import org.awaitility.Awaitility;
 import org.eclipse.jetty.server.LowResourceMonitor;
 import org.eclipse.jetty.server.ServerConnector;
 import org.junit.Rule;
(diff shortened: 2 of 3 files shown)


src/main/java/winstone/Launcher.java
```diff
@@ -229,8 +230,12 @@ private void writePortToFileIfNeeded() throws IOException {
                     Path portFile = Paths.get(portFileName);
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
-                    try (BufferedWriter writer = Files.newBufferedWriter(portFile, StandardCharsets.UTF_8)) {
+                    Path tmpPath = Files.createTempFile(portDir, "temp", "");
+                    try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
+                        Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
```
timja, line 236: should we handle the `AtomicMoveNotSupportedException` thrown here?

(may have been hidden due to the existing `IOException` already thrown)

src/test/java/winstone/HttpConnectorFactoryTest.java
```diff
@@ -49,6 +56,29 @@ public void testListenAddress() throws Exception {
...
+                }
+                try {
+                    Thread.sleep(1);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        };
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        Future<String> futurePort = executorService.submit(asyncVerifier);
```
timja, line 80: using something like this would result in cleaner code:
http://www.awaitility.org/

src/test/java/winstone/HttpConnectorFactoryTest.java
```diff
@@ -49,6 +56,29 @@ public void testListenAddress() throws Exception {
 
     @Test
     public void writePortInFile() throws Exception {
+        Callable<String> asyncVerifier = () -> {
+            System.out.println("Async Verifier started!");
+
+            while (true) {
```
olamy, line 62: if there is a bug and the file is not written we will rely on surefire timeout which can be a bit long.
at least we need some timeout but yup http://www.awaitility.org/ will be better.
see comment here https://github.com/jenkinsci/winstone/pull/233#issuecomment-1165099862

src/test/java/winstone/HttpConnectorFactoryTest.java
Pldi23: Thanks, applied in [8c54050](https://github.com/jenkinsci/winstone/pull/249/commits/8c540507e1e5797aac7189e09d1542bb672532a3)

src/main/java/winstone/Launcher.java
Pldi23: Error handling has been added in [a60732f](https://github.com/jenkinsci/winstone/pull/249/commits/a60732f8dc91a82a418e9773d3a123389f9d318c), please review and suggest any improvements if you'll have one.

src/main/java/winstone/Launcher.java
```diff
@@ -229,8 +231,17 @@ private void writePortToFileIfNeeded() throws IOException {
                     Path portFile = Paths.get(portFileName);
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
-                    try (BufferedWriter writer = Files.newBufferedWriter(portFile, StandardCharsets.UTF_8)) {
+                    Path tmpPath = Files.createTempFile(portDir, "temp", "");
+                    try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
+                        try {
+                            Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
```
Pldi23, line 238: I'm a little bit confused but in very rare case `HttpConnectorFactoryTest#writePortInFile` Test could still read empty file content, does anyone knows is `StandardCopyOption.ATOMIC_MOVE` absolutely guarantee atomic move?

src/test/java/winstone/HttpConnectorFactoryTest.java
```diff
@@ -49,6 +57,20 @@ public void testListenAddress() throws Exception {
...
+
+            Awaitility.await()
+                    .pollInterval(1, TimeUnit.MICROSECONDS)
+                    .atMost(5, TimeUnit.SECONDS)
+                    .until(() -> Files.exists(portFile));
+
+            return FileUtils.readFileToString(portFile.toFile(), StandardCharsets.UTF_8);
+        };
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        Future<String> futurePort = executorService.submit(parallelVerifier);
```
timja, line 72: what do we need this callable for?  can't we just await and once it's found the file read it into a string.

i.e. why do we need a future?

From my reading of the docs this doesn't look right, https://github.com/awaitility/awaitility/wiki/Usage

src/main/java/winstone/Launcher.java
timja: what line does it fail on? line 84 to 87 are not waiting for it to exist in the test from what I can tell

src/test/java/winstone/HttpConnectorFactoryTest.java
Pldi23: This Test tries to imitate that if parallel process trying to access the port the situation when `portFile` already exists, but `port` still not written can't happen, so we are using Future to get first read result after `portFile` detection and we need two threads running in parallel during test execution. I thought that `await` will always block the current thread until condition (I will re-read the doc, maybe I missed something).

src/main/java/winstone/Launcher.java
Pldi23: line 87, condition "Port value should not be empty in any time" (which comes from Future), fails very-very rare.

Pldi23: Flakiness was because `writer` wasn't auto closed. Fixed in [d4688e8](https://github.com/jenkinsci/winstone/pull/249/commits/d4688e8c29d7d2c1067badb750f5751a87ca05b5)

src/test/java/winstone/HttpConnectorFactoryTest.java
Pldi23: @timja maybe I'm missing something or we are misunderstanding each other, the goal of test changes is to verify that there is no time gap between `portFile` creation and `port` value write. Already existed asserts are fine and checks port value was written and they works as expected, but they don't check that write-process is atomic. So second Thread is used to hit portFile as often as possible and once file exists read the first value, later in `Future` we could check it is not empty (line 88). I don't see clear way to use  http://www.awaitility.org/ to imitate such scenario instead of second thread. Please correct me if I'm missing something.

timja: Ah I see I misunderstood :)

timja (approved):

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
                     Path portFile = Paths.get(portFileName);
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
-                    try (BufferedWriter writer = Files.newBufferedWriter(portFile, StandardCharsets.UTF_8)) {
+                    Path tmpPath = Files.createTempFile(portDir, "temp", "");
```
basil, line 234: Using `portDir` here is great! That ensures that the temporary file is created on the same file system as the target, so the atomic move is more likely to succeed. :thumbsup:

The use of `temp` as the prefix is a bit generic: why not use the name of the destination file as the prefix? That way the temporary file is more obviously connected to its destination which can make debugging easier.

Similarly, if you don't have a particular suffix in mind, I think it is more idiomatic to use `null` as the suffix, which gives us the default suffix of `.tmp`. I think it is better to stick with the default if we don't have a specific customization in mind.

```suggestion
                    Path tmpPath = Files.createTempFile(portDir, portFile.getFileName().toString(), null);
```

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
...
+                    try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
                     }
+
+                    try {
+                        Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
```
basil, line 240: In the first attempt, you are not using `StandardCopyOption.REPLACE_EXISTING`, but in the second attempt you are. I see no reason for this inconsistency: the semantics of atomicity are unrelated to the semantics of replacing an existing destination.

I would suggest making both use the same semantics: either `StandardCopyOption.REPLACE_EXISTING` in both attempts or in neither. Since the original code before your PR used `Files.newBufferedWriter`, which uses `StandardOpenOption.TRUNCATE_EXISTING`, the least risky change would be to retain the same semantics. So I suggest using `StandardCopyOption.REPLACE_EXISTING` in both attempts, which means adding it here:

```suggestion
                        Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);
```

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
...
+                    } catch (AtomicMoveNotSupportedException e) {
+                        Logger.logDirectMessage(Level.WARNING, "", "Atomic move does not supported", e);
```
basil, line 242: Elsewhere in this file, the `streamName` argument is always passed as `null` and `Logger.WARNING` is used rather than `Level.WARNING`, so for consistency I think we should do the same here. Also, I think the message could be improved to state that we are about to make another attempt:

```suggestion
                        Logger.logDirectMessage(Logger.WARNING, null, "Atomic move not supported. Falling back to non-atomic move.", e);
```

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
...
+                        Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
```
basil, line 243: If we encounter an exception on the second attempt, we will lose the stack trace from the first attempt, which is valuable for debugging. Better to attach it as a suppressed exception, which preserves the original stack trace from the first attempt:

```suggestion
                        try {
                            Files.move(tmpPath, portFile, StandardCopyOption.REPLACE_EXISTING);
                        } catch (IOException e2) {
                            e2.addSuppressed(e);
                            throw e2;
                        }
```

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
...
+                    } finally {
+                        tmpPath.toFile().deleteOnExit();
```
basil, line 245: The deletion here will always be a no-op, as the temporary path no longer exists at this point. So there is no reason to try and delete it. Suggest simply removing this unnecessary code:

```suggestion
```

src/test/java/winstone/HttpConnectorFactoryTest.java
```diff
@@ -49,6 +57,20 @@ public void testListenAddress() throws Exception {
...
+        Callable<String> parallelVerifier = () -> {
+            System.out.println("Async Verifier started!");
+            Path portFile = Paths.get(tmp.getRoot().getAbsolutePath(), "subdir/jenkins.port");
+
+            Awaitility.await()
+                    .pollInterval(1, TimeUnit.MICROSECONDS)
+                    .atMost(5, TimeUnit.SECONDS)
+                    .until(() -> Files.exists(portFile));
+
+            return FileUtils.readFileToString(portFile.toFile(), StandardCharsets.UTF_8);
```
basil, line 69: This repository is using Java 11, so no need for a third-party library anymore!

```suggestion
            return Files.readString(portFile, StandardCharsets.UTF_8);
```

src/test/java/winstone/HttpConnectorFactoryTest.java
```diff
@@ -63,6 +85,7 @@ public void writePortInFile() throws Exception {
                 String portInFile = reader.readLine();
                 assertEquals(Integer.toString(port), portInFile);
             }
+            assertFalse("Port value should not be empty in any time", futurePort.get().isEmpty());
```
basil, line 88: Optional feedback for your consideration: the `try` block above could be simplified to:

```java
assertEquals(Integer.toString(port), Files.readString(portFile, StandardCharsets.UTF_8));
```

If you take my suggestion above to use `Files.readString`, this would be consistent. However, not a blocking comment, since this is pre-existing code you did not write.

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
                     Path portFile = Paths.get(portFileName);
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
-                    try (BufferedWriter writer = Files.newBufferedWriter(portFile, StandardCharsets.UTF_8)) {
+                    Path tmpPath = Files.createTempFile(portDir, "temp", "");
+                    try (BufferedWriter writer = Files.newBufferedWriter(tmpPath, StandardCharsets.UTF_8)) {
                         writer.write(Integer.toString(port));
                     }
```
basil, line 237: Optional feedback for your consideration: This repository is using Java 11, so we have access to new `Files` methods! That allows us to reduce these three lines to one, which makes it easier to quickly read the code:

```suggestion
                    Files.writeString(tmpPath, Integer.toString(port));
```

Not a blocking comment, since the old `BufferedWriter` approach is pre-existing code that you did not write.

basil (commented):
Nice job! Looks good overall. I have a few comments on the implementation, but once those are addressed I should be ready to approve, merge, and release this.

src/test/java/winstone/HttpConnectorFactoryTest.java
basil: > we need two threads running in parallel during test execution

The race in the original code is not between two readers of the file but rather between the writer of the file and a reader of the file. The test as modified in this PR is not actually testing this scenario because it is always reading the file after the writer is finished.

To make the test actually test the scenario that failed in production you would need to do this:

```java
diff --git a/src/main/java/winstone/Launcher.java b/src/main/java/winstone/Launcher.java
index 0a8485b..bc3dba6 100644
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -230,6 +230,11 @@ public class Launcher implements Runnable {
                     Path portDir = portFile.getParent();
                     Files.createDirectories(portDir);
                     try (BufferedWriter writer = Files.newBufferedWriter(portFile, StandardCharsets.UTF_8)) {
+                        try {
+                            Thread.sleep(5000L);
+                        } catch (InterruptedException e) {
+                            // ignore
+                        }
                         writer.write(Integer.toString(port));
                     }
                 } else {
@@ -49,6 +58,7 @@ public class HttpConnectorFactoryTest extends AbstractWinstoneTest {
 
     @Test
     public void writePortInFile() throws Exception {
+        Future<Integer> futurePort = Executors.newSingleThreadExecutor().submit(() -> {
             Map<String,String> args = new HashMap<>();
             args.put("warfile", "target/test-classes/test.war");
             args.put("prefix", "/");
@@ -58,13 +68,20 @@ public class HttpConnectorFactoryTest extends AbstractWinstoneTest {
             System.setProperty(WINSTONE_PORT_FILE_NAME_PROPERTY, portFile.toString());
             try {
                 winstone = new Launcher(args);
-            int port = ((ServerConnector) winstone.server.getConnectors()[0]).getLocalPort();
-            try (BufferedReader reader = Files.newBufferedReader(portFile, StandardCharsets.UTF_8)) {
-                String portInFile = reader.readLine();
-                assertEquals(Integer.toString(port), portInFile);
-            }
+                return ((ServerConnector) winstone.server.getConnectors()[0]).getLocalPort();
             } finally {
                 System.clearProperty(WINSTONE_PORT_FILE_NAME_PROPERTY);
             }
+        });
+        Path portFile = Paths.get(tmp.getRoot().getAbsolutePath(), "subdir/jenkins.port");
+
+        Awaitility.await()
+                .pollInterval(1, TimeUnit.MICROSECONDS)
+                .atMost(5, TimeUnit.SECONDS)
+                .until(() -> Files.exists(portFile));
+
+        String portFileString = Files.readString(portFile, StandardCharsets.UTF_8);
+        assertEquals(Integer.toString(futurePort.get()), portFileString);
+        assertFalse("Port value should not be empty in any time", portFileString.isEmpty());
     }
 }
```

This adds a five-second sleep after truncating the file but before writing the value to expose the race. Then, from the test, we start the server asynchronously. The server truncates the file and starts sleeping. Now from the test we try to read the port and find that it is empty because the server has not written it yet (the server is still sleeping). This reproduces the original problem.

Obviously the five-second sleep should not be checked in. But if we are to change the test at all, we should change it as described above rather than as implemented in this PR. Without the sleep, the above change will result in the test failing as often as it did in production: rarely, but more often on systems with slow I/O.

src/main/java/winstone/Launcher.java
jtnord: > I would suggest making both use the same semantics: either StandardCopyOption.REPLACE_EXISTING in both attempts or in neither

as per the [javadoc ](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#move(java.nio.file.Path,java.nio.file.Path,java.nio.file.CopyOption...)) it would be incorrect to use both `REPLACE_EXISTING` and `ATOMIC_MOVE`, and lead to potentially bad assumptions when reading the code that existing files would be replaced (and that flag honoured) which may or may not be the case depending on your filesystem and operating system combination.

if the file is not expected to ever exist then I agree that REPLACE_EXISTING should not be needed for the non atomic case.

src/main/java/winstone/Launcher.java
```diff
@@ -229,9 +231,19 @@ private void writePortToFileIfNeeded() throws IOException {
...
+
+                    try {
+                        Files.move(tmpPath, portFile, StandardCopyOption.ATOMIC_MOVE);
+                    } catch (AtomicMoveNotSupportedException e) {
```
jtnord, line 241: You may also need to catch `IOException` in case the AtomicMOve operation s*is* supported, yet it is not a valid OS/filesystem combination that can overwrite an existing file.

See https://github.com/jenkinsci/jenkins/blob/0907d01d0a9e567b81dfca88ad854b5ef6dc9577/core/src/main/java/hudson/util/AtomicFileWriter.java#L193-L224 for some battle tested code.

If the file is not expected to exist this can be simplified somewhat but per the comment above from Basil and the code below wrt `REPLACE_EXISTING` it  am not sure if we expect it or not.

src/main/java/winstone/Launcher.java
basil: > as per the [javadoc ](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#move(java.nio.file.Path,java.nio.file.Path,java.nio.file.CopyOption...))it would be incorrect to use both REPLACE_EXISTING and ATOMIC_MOVE

I disagree with this. I see no evidence that it is incorrect to use both `REPLACE_EXISTING` and `ATOMIC_MOVE`.

src/main/java/winstone/Launcher.java
basil: YAGNI I think. While I don't doubt that there were some truly bizarre edge cases that led to the battle-tested code in core's `AtomicFileWriter`, I think that is overkill in this case and that it would be preferable to keep it simple and not deal with too many pathological cases other than the usual `AtomicMoveNotSupportedException`.

src/main/java/winstone/Launcher.java
basil: Oh I think I see what you mean: "all other options are ignored" when using `REPLACE_EXISTING`. So yes, it is redundant to specify both, and `ATOMIC_MOVE` is sufficient.

jtnord: Yes exactly that, sorry I was not clear.

src/main/java/winstone/Launcher.java
Pldi23: remove redundant deleteOnExit

src/test/java/winstone/HttpConnectorFactoryTest.java
Pldi23: resolved in [586feef](https://github.com/jenkinsci/winstone/pull/249/commits/586feeff71811269c7bc802b44b0b2fb760a4cfb)

src/test/java/winstone/HttpConnectorFactoryTest.java
Pldi23: I made changes based on your idea, please see [586feef](https://github.com/jenkinsci/winstone/pull/249/commits/586feeff71811269c7bc802b44b0b2fb760a4cfb)

basil (approved):
This PR is now ready for merge. We will merge it after approximately 24 hours if there is no negative feedback. Please see the [merge process documentation](https://github.com/jenkinsci/jenkins/blob/master/docs/MAINTAINERS.adoc#merge-process) for more information about the merge process. Thanks!

basil:
https://github.com/jenkinsci/winstone/releases/tag/winstone-6.2

PR #253, opened 2022-08-10
Improve comment and error message

basil:
Improve the comment from #246 to more accurately reflect what the code is doing, and improve the error message to state the name of the problematic entry when the condition is encountered.

CC @JLLeitschuh @daniel-beck

### DIFF
--- a/src/main/java/winstone/HostConfiguration.java
+++ b/src/main/java/winstone/HostConfiguration.java
@@ -281,9 +281,9 @@ public class HostConfiguration {
                     // If archive date is newer than unzipped file, overwrite
                     File outFile = new File(unzippedDir, elemName);
 
-                    // Fix Zip Slip vulnerability
+                    // Protect against Zip Slip attack
                     if (!outFile.toPath().normalize().startsWith(unzippedDir.toPath().normalize())) {
-                        throw new IOException("Bad zip entry");
+                        throw new IOException("Bad zip entry: " + elemName);
                     }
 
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

JLLeitschuh (approved):
Looks good to me!

basil:
I'll wait a day or so to see if @daniel-beck has any feedback about the phrasing of the code comment prior to merging.

src/main/java/winstone/HostConfiguration.java
```diff
@@ -281,9 +281,9 @@ protected File getWebRoot(File requestedWebroot, File warfile) throws IOExceptio
                     // If archive date is newer than unzipped file, overwrite
                     File outFile = new File(unzippedDir, elemName);
 
-                    // Fix Zip Slip vulnerability
+                    // Protect against Zip Slip attack
```
daniel-beck, line 284: This suggests there's a real attack possible when that's already slightly misleading.

```suggestion
                    // Disallow unzipping files outside the target dir)
```

daniel-beck (commented):
No strong opinion either way.

FTR I've already rephrased https://github.com/jenkinsci/winstone/releases/tag/winstone-6.2 based on the conversation in the previous PR, as it's not a vulnerability for lack of impact: there is no increase of access, privileges gained, or other negative effect from the previous behavior.

basil (commented):
Thanks for the feedback, Daniel!

PR #277, opened 2022-09-18
Migrate from HttpUnit to native Java Platform functionality

basil:
### Problem

The test suite has a dependency on [HttpUnit](http://httpunit.sourceforge.net/), last released in 2008 and which in turn depends on an obsolete version of Rhino. This does not seem sustainable.

### Solution

Migrate to the native HTTP client offered in Java 9+.

## Implementation

Note that `testKeepAliveConnection` appears to have never been completed; in practice, the `WebImage[] img = wresp1.getImages()` line was always returning an empty array, making this test equivalent to `testSimpleConnection`. The commented out lines also appear to indicate that the attempt to finish this test was abandoned. Since this test never worked to begin with and duplicated an existing test, I simply deleted it.

### Testing done

`LoadTest` appears to have been completely broken before this PR, but as of this PR it now works:

```
Sep 18, 2022 2:51:08 PM winstone.Logger logInternal
INFO: Load test initialised with properties: URL=http://127.0.0.1, KeepAlives=true, StartThreads=20, EndThreads=1000, StepSize=20, StepPeriod=5000, GracePeriod=5000
Sep 18, 2022 2:51:18 PM winstone.Logger logInternal
INFO: n=20, success=100, error=0, averageTime=99ms
Sep 18, 2022 2:51:28 PM winstone.Logger logInternal
INFO: n=40, success=199, error=1, averageTime=18ms
Sep 18, 2022 2:51:38 PM winstone.Logger logInternal
INFO: n=60, success=298, error=2, averageTime=18ms
```

### DIFF
--- a/src/test/java/winstone/testCase/HttpConnectorTest.java
+++ b/src/test/java/winstone/testCase/HttpConnectorTest.java
@@ -6,8 +6,15 @@
  */
 package winstone.testCase;
 
+import static org.junit.Assert.assertEquals;
+
 import java.io.IOException;
-import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
 import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.Map;
@@ -16,16 +23,8 @@ import junit.framework.Test;
 import junit.framework.TestCase;
 import junit.framework.TestSuite;
 
-import org.xml.sax.SAXException;
-
 import winstone.Launcher;
 import winstone.Logger;
-
-import com.meterware.httpunit.GetMethodWebRequest;
-import com.meterware.httpunit.WebConversation;
-import com.meterware.httpunit.WebImage;
-import com.meterware.httpunit.WebRequest;
-import com.meterware.httpunit.WebResponse;
 
 /**
  * Test case for the Http Connector to Winstone. Simulates a simple connect and
@@ -49,8 +48,8 @@ public class HttpConnectorTest extends TestCase {
     /**
      * Test the simple case of connecting, retrieving and disconnecting
      */
-    public void testSimpleConnection() throws IOException, SAXException,
-            InterruptedException {
+    public void testSimpleConnection()
+            throws URISyntaxException, IOException, InterruptedException {
         // Initialise container
         Map<String,String> args = new HashMap<>();
         args.put("webroot", "target/testwebapp");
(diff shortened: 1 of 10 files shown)


src/test/java/winstone/testCase/HttpConnectorTest.java
```diff
@@ -64,58 +63,23 @@ public void testSimpleConnection() throws IOException, SAXException,
...
-        // Thread.sleep(2000);
-        // WebResponse wresp2 = wc.getResponse(wreq);
-        // Thread.sleep(2000);
-        //WebResponse wresp3 = wc.getResponse(wreq);
-        InputStream content = wresp1.getInputStream();
-        assertTrue("Loading CountRequestsServlet + child images", content
-                .available() > 0);
-        content.close();
+        HttpRequest request =
+                HttpRequest.newBuilder(new URI("http://localhost:10003/examples/CountRequestsServlet")).GET().build();
```
olamy, line 67: while you're doing some cleaning. maybe using a random port here would be good as well.
same as done in `ArgumentsRealmTest`

src/test/java/winstone/testCase/HttpConnectorTest.java
```diff
@@ -129,12 +93,14 @@ public void testWriteAfterServlet() throws IOException,
...
 
         // Check for a simple connection
-        WebConversation wc = new WebConversation();
-        WebRequest wreq = new GetMethodWebRequest(
-                "http://localhost:10005/examples/TestWriteAfterServlet");
-        WebResponse wresp1 = wc.getResponse(wreq);
-        Logger.logDirectMessage(Logger.INFO, "log", "Output: " + wresp1.getText(), null);
-        assertTrue(wresp1.getText().endsWith("Hello"));
+        HttpRequest request =
+                HttpRequest.newBuilder(new URI("http://localhost:10005/examples/TestWriteAfterServlet")).GET().build();
```
olamy, line 97: while you're doing some cleaning. maybe using a random port here would be good as well.
same as done in ArgumentsRealmTest

olamy (approved):

basil:
> while you're doing some cleaning

I agree that is a good idea, but I am not willing to take additional requests while I am here. What I have offered is all I am willing to offer.

src/test/java/winstone/testCase/HttpConnectorTest.java
basil: Not interested

src/test/java/winstone/testCase/HttpConnectorTest.java
basil: Not interested

timja (approved):

PR #282, opened 2022-09-19
[JENKINS-69638] HTTP/2 broken after Jetty 10 upgrade

basil:
### Problem

Running Winstone in HTTP/2 mode fails with

```
WARNING: Failed to start ALPN
java.lang.IllegalStateException: No Server ALPNProcessors!
        at org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory.<init>(ALPNServerConnectionFactory.java:47)
        at winstone.Http2ConnectorFactory.start(Http2ConnectorFactory.java:79)
        at winstone.Launcher.spawnListener(Launcher.java:269)
        at winstone.Launcher.<init>(Launcher.java:185)
```

I reproduced this error in an integration test.

### Solution

As of https://github.com/eclipse/jetty.project/issues/215 and https://github.com/eclipse/jetty.project/issues/2948 (specifically, eclipse/jetty.project@0303ae4aafe20c6177a834320da90cb89944c01b, eclipse/jetty.project@8b14cbe5f873d7b930ded9872675520d2cf2cf90, eclipse/jetty.project@1fe6c92ee9cce80c180eafb792435b7b6af3d86c, and eclipse/jetty.project@31a9b6f2e84a23f9bdffae2cacff88ab8ea1fc4b), the `alpn-api` dependency is no longer used. In its place is a new `jetty-alpn-java-server` based on Java 11 and later. Including that correct dependency fixes the problem.

### Implementation

Added a new test that now passes with the fix. The changes to the Surefire configuration in `pom.xml` are explained in https://github.com/jenkinsci/winstone/pull/281.

### Testing done

The new test fails without the changes in `src/main` and `pom.xml` and passes with them. I also tested this end-to-end by running Jenkins with these changes and successfully running `curl --http2 127.0.0.1:80` (previously, this was failing).

PR #53, opened 2018-08-16
[JENKINS-52358] fix broken http/2 support

olamy:
add an extra lib directory option to be able to add alpn processor

Signed-off-by: olivier lamy <olamy@apache.org>

PR #290, opened 2022-09-28
Deprecate `--extraLibFolder` option

basil:
The `--extraLibFolder` option was originally added in #53 for HTTP/2 support. While potential secondary use cases remain, the primary use case no longer exists as of #282. Deprecating this functionality so that it can hopefully be removed in the future to ease the burden of maintaining this component.


NotMyFault (approved):

timja (approved):

olamy:
Does this might prevent users to add some extra jars to the jetty classpath such jetty sessions manager? cc @Vlatombe

basil:
If for some reason we need additional Jetty functionality in the future, like a session manager, we can always bundle it with Winstone.

PR #295, opened 2022-10-12
De-duplicate file-based logging implementations

basil:
See https://github.com/jenkinsci/jenkins/pull/7256 for the PR description.


src/main/java/winstone/Launcher.java
```diff
@@ -358,6 +359,8 @@ public boolean isRunning() {
      * listener thread. For now, just shut it down with a control-C.
      */
     public static void main(String[] argv) throws IOException {
+        Map<String, String> args = getArgsFromCommandLine(argv);
```
basil, line 362: Critical that we parse `--logfile` and redirect the streams _before_ initializing JUL, as explained in the block comment in the `Logger` class.

src/main/java/winstone/Launcher.java
```diff
@@ -461,27 +462,23 @@ protected static void deployEmbeddedWarfile(Map<String, String> args) throws IOE
...
-            logStream = System.err;
-            logCharset = Charset.defaultCharset();
-        } else {
-            logStream = System.out;
-            logCharset = Charset.defaultCharset();
+            OutputStream outputStream = Files.newOutputStream(logPath, StandardOpenOption.CREATE, StandardOpenOption.APPEND);
+            // TODO: Ideally, should change this to UTF-8, but this could cause problems for Windows users when appending to existing logs.
+            PrintStream printStream = new PrintStream(outputStream, false, Charset.defaultCharset());
+            System.setOut(printStream);
+            System.setErr(printStream);
```
basil, line 479: Key change in semantics: redirecting both standard streams when `--logfile` is specified. This was always the behavior in Jenkins' `--logfile` but I suspect it was not the case in (pre-Jetty) Winstone.

src/main/java/winstone/Logger.java
```diff
@@ -6,160 +6,61 @@
...
-                    stream = defaultStream;
-                }
-                return stream;
-            }
-        } else {
-            return defaultStream;
-        }
-
-    }
-
```
basil, line 157: All of this streams-related logic was dead code that was setting fields but never getting them. Deleting it makes it more clear what is going on in this class, which is really not very much other than delegating to JUL.

src/main/java/winstone/Logger.java
```diff
@@ -6,160 +6,61 @@
...
             boolean showThrowingThreadArg) {
         synchronized (semaphore) {
             if (!initialised) { // recheck in case we were blocking on another init
-                initialised = false;
-                LOGGER.setLevel(level);
-                namedStreams = new HashMap<>();
-//                nullStreams = new ArrayList();
-                initialised = true;
-                setStream(DEFAULT_STREAM, defaultStream, defaultStreamCharset);
+                java.util.logging.Logger.getLogger("").setLevel(level);
```
basil, line 52: Key change in semantics: setting the level on the root logger rather than just Winstone's logger. This enables the `--debug` argument to be mildly useful for Jenkins users.

src/main/java/winstone/cmdline/Option.java
```diff
@@ -203,6 +205,39 @@ public int get(Map<String, String> args, int defaultValue) {
...
+                    return Level.FINEST.intValue();
+                case 9:
+                    return Level.ALL.intValue();
+                case 5:
+                default:
+                    return Level.INFO.intValue();
+            }
+        }
+    }
+
```
basil, line 240: The mapping is a bit funky, but this was just extracted from existing code in Winstone and generalized to apply to both `Option` and `WinstoneControl`.

timja (approved):

PR #307, opened 2022-12-21
`--debug` should only control Winstone log level

basil:
This is a partial revert of the (unreleased) changes in #295. While trying to make sense of the semantics of the `--debug` option in #295 I realized that it was (then) misleading, as it appeared to set the log level for the root logger but was only setting it for the Winstone logger. I found this inconsistent with the `--logfile` option, which was setting the log file for all logs, so I modified `--debug` to set the log level for the root logger. But after thinking about that more, I think this is a bad idea and that the original behavior was better. I came to this conclusion after doing some more testing and realized that the default "All Jenkins Logs" log recorder prints the message "Log messages at a level lower than INFO are never recorded in the Jenkins log. Use log recorders to record these log messages" and requires users to create a custom log recorder to print more detailed messages, and if users attempt to do this then they are strongly warned that

> Loggers that record messages at FINE, FINER, or even FINEST can cause performance issues when many messages are being logged. More verbose log levels increase the likelihood of credentials being logged. It is recommended that you delete such loggers, or update them to a less verbose level, when you no longer need them.

With my previous change, nothing warns users from setting `--debug` to a high level in Winstone. Previously this had no effect other than on a handful of seldom-used Winstone log statements, but when extended to the root logger as in #295 it seems likely that this could be misused. So I think it is safer to maintain the older behavior, as inconsistent as it is internally within Winstone. I have clarified the documentation to make the old behavior less surprising though.


basil:
Updated the PR description of https://github.com/jenkinsci/jenkins/pull/7256 to reflect the reduced scope and re-did the testing described in the **Testing Done** section. Once this PR is approved, I will be ready to release Winstone 6.7 and https://github.com/jenkinsci/jenkins/pull/7256.

timja (approved):

PR #354, opened 2023-11-10
optional client certificate

### LINKED ISSUE #353: allow optional clientauth certificate
### What feature do you want to see added?

We have a plugin in Jenkins that allows us to login to Jenkins by reading the user from a client certificate, which is very convenient in our environment. We run Jenkins with tomcat where you can configure that a client certificate is optional. So when we access Jenkins via browser, the certificate is sent and we're automatically logged in, but also accessing the REST api with user/token works very well.
Currently there is the option `--httpsVerifyClient` which is a boolean. `true` means the client must always include a certificate, which makes it hard access the rest api with curl, `false` means a client certificate is never requested.
So it would be nice to have `--httpsVerifyClient` support three value: `true`, `false` and `optional` 
Jetty supports optional client certificates via `ssl.setWantClientAuth(true);`

### Upstream changes

_No response_

mawinter69:
allow to make the client verification via a certificate optional.

fix #353

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/winstone/AbstractSecuredConnectorFactory.java
+++ b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
@@ -128,7 +128,8 @@ public abstract class AbstractSecuredConnectorFactory implements ConnectorFactor
                     ssl.setWantClientAuth(true);
                     break;
                 default:
-                    break;
+                   ssl.setNeedClientAuth(false);
+                   break;
             }
             return ssl;
         } catch (Throwable err) {
--- a/src/main/java/winstone/AbstractSecuredConnectorFactory.java
+++ b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
@@ -118,16 +118,15 @@ public abstract class AbstractSecuredConnectorFactory implements ConnectorFactor
                         "HttpsListener.ExcludeCiphers", //
                         Arrays.asList(ssl.getExcludeCipherSuites()));
 
-            /*
-             * If true, request the client certificate ala "SSLVerifyClient require" Apache directive.
-             * If false, which is the default, don't do so.
-             * Technically speaking, there's the equivalent of "SSLVerifyClient optional", but IE doesn't
-             * recognize it and it always prompt the certificate chooser dialog box, so in practice
-             * it's useless.
-             * <p>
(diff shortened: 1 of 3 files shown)


mawinter69:
@basil @olamy 
Can you please take a look

mawinter69:
Friendly Reminder @basil @olamy

src/main/java/winstone/AbstractSecuredConnectorFactory.java
```diff
@@ -118,16 +119,17 @@ protected SslContextFactory.Server getSSLContext( Map<String, String> args) {
...
-             * <p>
-             * See http://hudson.361315.n4.nabble.com/winstone-container-and-ssl-td383501.html for this failure mode in IE.
-             */
-            ssl.setNeedClientAuth(Option.HTTPS_VERIFY_CLIENT.get(args));
+            switch (Option.HTTPS_VERIFY_CLIENT.get(args).toLowerCase(Locale.ROOT)) {
+                case "yes":
+                case "true":
+                    ssl.setNeedClientAuth(true);
+                    break;
+                case "optional":
```
olamy, line 127: this need some update in the help file here https://github.com/jenkinsci/winstone/blob/master/src/main/resources/winstone/LocalStrings.properties

olamy (changes requested):

src/main/java/winstone/AbstractSecuredConnectorFactory.java
mawinter69: So far this option is undocumented but I can add it

src/main/java/winstone/AbstractSecuredConnectorFactory.java
```diff
@@ -118,16 +119,17 @@ protected SslContextFactory.Server getSSLContext( Map<String, String> args) {
...
+                    ssl.setWantClientAuth(true);
+                    break;
+                default:
+                    break;
```
olamy, line 131: nit for readiness I would prefer something such rather than doing nothing 
 ```suggestion
                    ssl.setNeedClientAuth(false);
                    break;
```

olamy (changes requested):

olamy (approved):
LGTM

olamy:
@mawinter69 Thanks a lot for your contribution. And your patience! ;) 
PR merged and release done

PR #376, opened 2024-05-03
Remove missed references to `--hostsDir`

basil:
commit a60ab16e84e0aec3ac9282240246953ce9fea2a4 from over a decade ago removed support for `--hostsDir`, but it missed a few lingering references in the `README` and in a localization file.

### Testing done

`git grep -i hostsdir` and `git grep -i hosts.dir`


basil:
Expediting this PR, as it is a trivial code cleanup with no changelog entry, and as I have another PR which builds on this change.

PR #377, opened 2024-05-03
Remove support for multiple WAR files

basil:
Winstone contains 100 or so lines of code for a `--webappsDir` switch to run a directory full of WAR files, each with their own request logger which logs requests to a different file. There are no references to this in the `jenkinsci`, `jenkins-infra`, and `cloudbees` GitHub organizations. There is no test coverage for the feature. The single reference in Jira is [JENKINS-5307](https://issues.jenkins.io/browse/JENKINS-5307), from 2010. While I am aware of people running Jenkins as a WAR file in Tomcat alongside other WAR files, I am not aware of anyone using Winstone to run multiple WAR files. I think it is quite likely there are zero people using this feature.

This all is changing in Jetty 12, which now does request logging at a per-`Server` level, not a per-webapp level. Thus the current code would require significant changes to support the existing functionality. Since it has no test coverage, and since we think there are zero people using it, this PR removes the feature. That way, we won't have to port it to Jetty 12, and we also won't have to manually test the result.

### Proposed changelog and upgrade guide entries

The `--webappsDir` argument to run Winstone with a directory full of WAR files has been removed without replacement.

### Testing done

`mvn clean verify`

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,6 @@ To run a single war file:
 
     java -jar winstone.jar --warfile=<location of warfile> (+ other options)
 
-To run a directory full of war files:
-
-    java -jar winstone.jar --webappsDir=<location of webapps directory> (+ other options)
-
 To run locally exploded web archive:
 
     java -jar winstone.jar --webroot=<location of webroot> (+ other options)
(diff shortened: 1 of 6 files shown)


basil:
If we feel that removal without first deprecating this option is too aggressive, then I can file a separate PR to first deprecate it and then remove it in a few months. However, based on the research I provided in the PR description, I think it is highly likely there are zero people using this feature and that it can be treated as dead code and removed without a prior deprecation period.

MarkEWaite:
> If we feel that removal without first deprecating this option is too aggressive, then I can file a separate PR to first deprecate it and then remove it in a few months. However, based on the research I provided in the PR description, I think it is highly likely there are zero people using this feature and that it can be treated as dead code and removed without a prior deprecation period.

I think it should be removed without deprecation.

MarkEWaite (approved):

PR #383, opened 2024-05-08
[JENKINS-73126] Upgrade Winstone from Jetty 10.x to 12.x (EE 8)

basil:
Replace dependencies with EE 8 versions in `pom.xml` and update all relevant imports.

### Testing done

Full PCT/ATH


MarkEWaite (approved):

src/main/java/winstone/HostConfiguration.java
```diff
@@ -55,7 +52,6 @@ public class HostConfiguration {
     private Map<String, String> args;
     private Map<String, WebAppContext> webapps;
     private ClassLoader commonLibCL;
-    private MimeTypes mimeTypes = new MimeTypes();
```
basil, line 58: Note that we are no longer replacing the default MIME types, but appending to them. As a result, we have deleted any MIME mappings that are already present in upstream Jetty and added a test to ensure that we are only adding to the existing MIME types rather than attempting to replace them.

src/test/java/winstone/HostConfigurationTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.Test;
+import org.junit.rules.ErrorCollector;
+
+public class HostConfigurationTest {
+
+    @Rule
+    public ErrorCollector errors = new ErrorCollector();
+
+    @Test
+    public void mimeTypes() throws IOException {
```
basil, line 18: This test ensures that we are only ever adding to the Jetty default MIME types, never replacing them. When upstream Jetty has a MIME type, we want to use it whenever possible. This is not only so that we have less code to maintain, but also so that we avoid bugs like JENKINS-73381.

BorisYaoA (approved):

alecharp (approved):
Looks great.

Kevin-CB (approved):
Security-wise, LGTM

PR #393, opened 2024-06-25
Add automated test for JENKINS-73285

basil:
Add a minimal reproducible test case for [JENKINS-73285](https://issues.jenkins.io/browse/JENKINS-73285). I plan to submit an upstream Jetty issue report for this, using this test as an example. The test passes on Jetty 10 and starts failing on Jetty 12 EE 8.

### Testing done

Ran the test successfully on Jetty 10. On Jetty 12 EE 8 the test fails with the stack trace reported in the issue.

### DIFF
--- a/src/test/java/winstone/FormSubmissionTest.java
+++ b/src/test/java/winstone/FormSubmissionTest.java
@@ -49,4 +49,24 @@ public class FormSubmissionTest extends AbstractWinstoneTest {
                     response.body());
         }
     }
+
+    @Issue("JENKINS-73285")
+    @Test
+    public void duplicateKeys() throws Exception {
+        Map<String, String> args = new HashMap<>();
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

olamy:
Looks to be more an issue with the configuration.
https://github.com/jenkinsci/winstone/blob/0bedfa07741b1b619e6211194cad106f5a940559/src/main/java/winstone/HostConfiguration.java#L180

default value -1 (unlimited) is never configured because of this test 
```
if (maxParameterCount > 0) {
```
so this turn into default max value, If Jenkins really needs such high value it's probably better to fix it when configuring Jetty in Winstone code.

olamy:
PR https://github.com/jenkinsci/winstone/pull/394

PR #397, opened 2024-07-01
[JENKINS-73381] Downloading `tar.gz` artifacts in Firefox is broken

basil:
### Problem

When downloading `tar.gz` artifacts with Firefox, which sets `Accept-Encoding: gzip, deflate, br, zstd`, the downloaded file is broken and has a different size than when downloaded with other browsers like Edge or Chrome.

### Evaluation

`GzipHandler` incorrectly kicks in, double-gzipping the file. Normally `GzipHandler` ignores the `application/gzip` MIME type, but we were incorrectly sending the MIME type `application/x-gtar`.

### Solution

Send the correct MIME type by bringing `mime.properties` closer to upstream Jetty. Note that the explicit `tar.gz` mapping is only needed for Jetty 10 and must be removed pending the adoption of jetty/jetty.project#8919.

### Testing done

- Reproduced the issue as described in the ticket on Firefox. Also confirmed the issue was not visible on Chrome. After this PR, both Firefox and Chrome work.
- Added new unit test that fails before this PR and passes after.

### Submitter checklist


src/main/resources/winstone/mime.properties
```diff
@@ -776,7 +778,7 @@ texinfo=application/x-texinfo
 text=text/plain
 tfi=application/thraud+xml
 tfm=application/x-tex-tfm
-tgz=application/x-gzip
+tgz=application/x-gtar
```
StefanVK, line 781: Shouldn't tgz have the same mime type as tar.gz?

Now I have different sizes when downloading a file as .tar.gz than when downloading it as .tgz

basil: We're not doing anything different from upstream Jetty here:

https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-core/jetty-http/src/main/resources/org/eclipse/jetty/http/mime.properties#L163

If there is a bug here, please report it to the upstream Jetty project and then we will pull it in automatically as part of a future Jetty upgrade.

felfert: Initially, I had the same idea, however after further evaluation, instead of having the developers
hardcoding custom mimetypes, I found an easier way to fix arbitrary
custom mimetypes: Just add `--mimeTypes=tgz=application/gzip` to the jenkins commandline.
When using systemd, you can specify any additional cmdline options for jenkins by specifying the
following in your jenkins override file:
`Environment="JENKINS_OPTS=--mimeTypes=tgz=application/gzip"`
This fixed download of .tgz artifacts for me.

Just my 2cent

StefanVK: @felfert thank your for this workaround as well.
For others who may stumble across this issue when searching for the problem:
Basil has since [reported the issue to jetty](https://github.com/jetty/jetty.project/issues/12397) himself and it's already [fixed in jetty](https://github.com/jetty/jetty.project/pull/12414).
Here's a [jenkins issue](https://issues.jenkins.io/browse/JENKINS-73942) for the problem. The fix will eventually make its way into jenkins. Fortunately there are also workarounds like the suggestion from @felfert until then.

PR #409, opened 2024-08-28
[JENKINS-73616] Log spam from `AnnotationParser#parseClass`

basil:
### Context

See [JENKINS-73616](https://issues.jenkins.io/browse/JENKINS-73616).

### Problem

Users have complained that there are new warnings emitted on Jenkins startup.

### Evaluation

These duplicates are known to the project and ignored in https://github.com/jenkinsci/jenkins/blob/c5c69abd2d346a694c29b3e5f5d3991d94c38556/test/src/test/java/jenkins/ClassPathTest.java because we cannot upgrade `jansi` without also upgrading Groovy as described in https://github.com/jenkinsci/jenkins/pull/5184#pullrequestreview-570261350 (though I doubt anyone has actually tried?).

### Solution

Since these warnings are a false positive, turn up the log level for this class from the default level to `SEVERE` so that this log statement is not printed.

### Testing done

Reproduced the issue with `java -jar jenkins.war` and could no longer reproduce after this PR.

### DIFF
--- a/src/main/java/winstone/Launcher.java
+++ b/src/main/java/winstone/Launcher.java
@@ -39,6 +39,7 @@ import java.util.logging.ConsoleHandler;
 import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
+import org.eclipse.jetty.ee8.annotations.AnnotationParser;
 import org.eclipse.jetty.jmx.MBeanContainer;
 import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.LowResourceMonitor;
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
Thanks very much!

I had tried something very close to this (using literal string as argument to getLogger instead of import) and had seen the `LG_LOST_LOGGER_DUE_TO_WEAK_REFERENCE` spotbugs warning yet when I ran with my changes, it did not suppress the log messages .  Today, I did the same thing today and it works.  Your technique with the import is much better, since it assures that the class is available at compile time.  I must have been missing something yesterday when I was experimenting, but I don't know what it was.

I ran without your changes and I see the log messages.  I ran with your changes and the log messages are not displayed  Thanks very much!

PR #417, opened 2024-10-10
Tolerate missing keystore password, for example for tests

Vlatombe:
Discovered while working on https://github.com/jenkinsci/jenkins-test-harness/pull/858

### Submitter checklist

### DIFF
--- a/src/main/java/winstone/AbstractSecuredConnectorFactory.java
+++ b/src/main/java/winstone/AbstractSecuredConnectorFactory.java
@@ -49,7 +49,7 @@ public abstract class AbstractSecuredConnectorFactory implements ConnectorFactor
 
                 keystore = KeyStore.getInstance(KeyStore.getDefaultType());
                 try (InputStream inputStream = new FileInputStream(keyStore)) {
-                    keystore.load(inputStream, this.keystorePassword.toCharArray());
+                    keystore.load(inputStream, this.keystorePassword == null ? null : this.keystorePassword.toCharArray());
                 }
             } else {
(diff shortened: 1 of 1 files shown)


jglick:
Does it work to just pass a `new char[0]` or is that considered distinct from “no password”?

jglick:
https://github.com/jenkinsci/winstone/blob/3afbe6c64f758d26e5784070cd425d1a12653581/src/main/resources/winstone/LocalStrings.properties#L66 so does this mean that not passing that parameter caused Winstone to crash with an NPE? Has this been broken forever? Seems like something that would get noticed quickly.

Vlatombe:
> not passing that parameter caused Winstone to crash with an NPE?

yup

jglick:
If that is a regression, it sounds like an `lts-candidate`.

Vlatombe:
https://github.com/jenkinsci/winstone/pull/16 then https://github.com/jenkinsci/winstone/commit/fceac88d0928a901e229c46771eefe514ff47eda.

jglick:
Huh. I guess everyone has been forced to set a password for the past decade?! Seems weird to me, because it adds little apparent security: you are making the password accessible to the machine running Jenkins (in fact part of the `/proc` command line, and the JVM’s reported args, and presumably in some service wrapper config), so it is not as if you are making it harder for someone who steals the keystore off the filesystem to use it.

Vlatombe:
Doesn't even seem to be possible to create a keystore using `keytool` without setting a password. Programatically this seems possible.

PR #533, opened 2026-03-25
Add support for H2C protocol, so client who wants to upgrade from http 1.1 to h2c can now, obviously not supported in https, (h2c is not used over SSL and HTTP/2 over TLS uses ALPN)

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/winstone/ServerConnectorBuilder.java
+++ b/src/main/java/winstone/ServerConnectorBuilder.java
@@ -3,6 +3,7 @@ package winstone;
 import java.nio.file.Path;
 import org.eclipse.jetty.http.HttpCompliance;
 import org.eclipse.jetty.http.UriCompliance;
+import org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory;
 import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.ForwardedRequestCustomizer;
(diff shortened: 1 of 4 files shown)


src/main/java/winstone/ServerConnectorBuilder.java
```diff
@@ -116,7 +117,13 @@ public Connector build() {
...
                 sc = new ServerConnector(server, acceptors, selectors, sslContextFactory);
             } else {
-                sc = new ServerConnector(server, acceptors, selectors);
+                HttpConfiguration hc = new HttpConfiguration();
+                sc = new ServerConnector(
+                        server,
+                        acceptors,
+                        selectors,
+                        new HttpConnectionFactory(hc),
+                        new HTTP2CServerConnectionFactory(hc));
```
olamy, line 126: should we have this configurable?

jglick: Probably.

jglick:
Does not address the problem outlined in https://github.com/jetty/jetty.project/pull/14730#issuecomment-4126543834.

